PR #9, opened 2017-05-12
Check with multiple versions of R on travis

jennybc:
I think it's a good idea to start checking with multiple versions of R on travis. Also we'll only build vignettes, run examples, and run covr on r-release.


jennybc:
@LucyMcGowan Have a look and merge at your convenience.

PR #13, opened 2017-05-12
Add `drive_list()` = renamed and upgraded `drive_ls()`

LucyMcGowan:
I've updated `drive_ls()` to look more like `list.files`
* a small demo is shown [here](https://github.com/tidyverse/googledrive/blob/drive_list/internal/07_drive_ls_demo.md).


LucyMcGowan:
@jennybc let me know what you think. This ultimately changes the `drive_ls()` function to mimic the `list.files` function. Basically I:
* created a file of helper functions ([`drive_folders.R`](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_folders.R)) that:
1. pulls the most recent 100 folders (maybe we'll have to adjust the `pageSize` here in case someone has LOTS of folders?)
2. figures out what the `root` folder is
3. matches the names the user inputs with these folder names
4. iteratively climbs backwards through the files to figure out the 🌳 
* the bulk of the change is [here](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_ls.R#L45-L70). I'm going to step through the function (let me know if you think the logic here checks out)
    * I basically pull the `path` the user inputs apart into folders and check whether those folders exist & grab their ids (from my helper functions) [here](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_ls.R#L47-L48)
    * I check if the file at the end of the path is really in the root folder [here](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_ls.R#L50-L52)
    * I check if the folder is unique in the root [here](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_ls.R#L54-L57). This may not really be necessary, but it is weird that google lets you have 2 folders named the same thing in the same folder...
    * I have a while loop that uses [this](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_folders.R#L44-L50) function which attempts to match each folder to a parent until there is only one folder left in the most sub-directory [here](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_ls.R#L59-L66)
*I've tested this only **very** briefly, so it needs more to make sure it's doing what I really think it's doing. I would love if you could break it!*
    * I then query based on the parent in the most sub-directory [here](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_ls.R#L122-L135).

Thank you!

jennybc:
I've just started to take a look at this. But here's one thing I'm already wondering:

Since plain old `drive_ls()` will show you things at all levels of the hierarchy (such as it is), I wonder if we should indicate in any way, for example, that a file does not live in "root".

Thinking more ... does it even matter? I could still ask for that file by name *without the path* if a felt like it. Right?

It's really true that folders on Drive do not mean what they mean in your local filesystem. It really is more like a tag.

No response necessary (or at least not at this point), but this probably should inform the interface.

R/drive_ls.R
```diff
@@ -26,51 +28,161 @@
...
+  }
+  len <- length(my_folder_ids)
+
+  while (nrow(my_folder_ids[[len - 1]]) != 1){
+    len <- length(my_folder_ids)
+    f1 <- my_folder_ids[1:(len - 1)]
+    f2 <- my_folder_ids[2:len]
+    my_folder_ids <- purrr::map2(f1, f2, folder_check)
+  }
+  my_folder_ids[[len - 1]]$id
```
jennybc: Yes, I do not succeed with your demo. But if I change this to `len`, I do.

jennybc:
You should write some tests, even though we'll need a way to easily skip them until we get more infrastructure in place. But you should be testing regularly as you go through the revision. I'll make a PR or push to this branch a low-tech solution for skipping, for now.

jennybc:
You might have to expand lots of comments that show as "outdated" just because I pushed some stuff since I made the comment.

R/drive_ls.R
LucyMcGowan: Oh hmm on my end I need `len-1` because I calculate `len` as the length of the list prior to performing the `map2` function, and after that function the length should be `len-1`, so to grab the last one it has to be `len-1`.

R/drive_ls.R
```diff
@@ -26,51 +28,161 @@
...
+  if (!sum(my_folder_ids[[1]]$root) == 1){
+    spf("It seems you have more than one folder named '%s' in your 'My Drive' (root) directory.", folders[1])
+  }
+
+  if (length(my_folder_ids) == 1){
+    return(my_folder_ids[[1]]$id[my_folder_ids[[1]]$root == TRUE])
+  }
+  len <- length(my_folder_ids)
+
+  while (nrow(my_folder_ids[[len - 1]]) != 1){
```
LucyMcGowan: The problem I'm facing is the iterative nature (not knowing how many to write without a `while`) because if you input `foo/bar/baz` I need to check 2 things, is there a folder named `baz` in `bar` and then if so is that same folder `bar` in `foo`, but if I have `foo/bar/baz/yo` then I need to check 3 things, is `yo` in `baz` and is `baz` in `bar` and is `bar` in `foo`. Other than using a `for` loop (are those better?) I couldn't think of a simpler way (but very open to suggestions!)

internal/07_drive_ls_demo.md
```diff
@@ -0,0 +1,75 @@
...
+``` r
+drive_ls(path = "foo/bar/baz")
+```
+
+    ## # A tibble: 2 × 4
+    ##      name     type                                           id
+    ##     <chr>    <chr>                                        <chr>
+    ## 1      yo   folder                 0Bw9rJumZU4vESVNLemFsQ3c1bGc
+    ## 2 my_file document 1fihkmyC76HCxvjYq6bVC4N2vG90G25FzfTBI6xe9sXg
+    ## # ... with 1 more variables: gfile <list>
```
LucyMcGowan: that's strange - do we have the same file structure on drive? I just reran it with the same function (prior to your changes) & got above. I have in my drive
`foo`
L`bar`
L`baz`
L`yo` `my_file`

R/drive_ls.R
```diff
@@ -26,51 +28,161 @@
...
+                     pattern = NULL,
+                     ...,
+                     fixed = FALSE,
+                     verbose = TRUE){
+
+  folder <- NULL
 
-  request <- build_drive_ls(...)
+  if (!is.null(path)){
+    folder <- climb_folders(path = path)
```
LucyMcGowan: yes! leafmost is must better description.

internal/07_drive_ls_demo.md
LucyMcGowan: do you get `baz` as the output of `climb_folders`? if so that would be the correct output, since that is the parent we are interested in.

LucyMcGowan:
thank you so much for taking the time to go through all of this! your comments are quite helpful!

internal/07_drive_ls_demo.md
LucyMcGowan: (or it is *very* possible that I have something funky going on in mine that I am missing)

jennybc:
Thinking at a 40,000 level about how `drive_list()` should handle path as input:

If someone passes a path, you need to split into folders.

Retrieve all folders (for some definition of all). ⬅️ should be another call to `drive_list()`! Or maybe just retrieve folders with names matching the pieces from step one?

Link up the folder names from step one in the listing from step two. Make sure they have the parent-child, parent-child, parent-child relationship that the original path implies. Error for funky stuff.

Take the folder that is farthest from root and list everything below that. ⬅️ should be another call to `drive_list()`!

LucyMcGowan:
would this cause weird recursion things? calling a function from within the same function?

jennybc:
> would this cause weird recursion things? calling a function from within the same function?

No, it's the right thing to do here.

You're really just taking the user's call to `drive_ls()` and processing its arguments, by making an intermediate call to `drive_ls()` with `path` re-worded for the API, in order to make the "correct" call: a call to `drive_ls()` with the query `'1234567' in parents` where '1234567' is the id of the lowest directory in the given path.

jennybc:
Sketch:

user says: `drive_list(path = "foo/bar/baz")`

internal work:
path is not NULL? ok foo/bar/baz --> foo, bar, baz
`drive_list( <get me all folders>)` or `drive_list( <get me all folders named foo, bar, or baz>)`
save ids: foo, bar, baz --> folder ids 123, 345, 567
make sure foo/123 has root as parent, bar/345 has foo/123 as parent, baz/567 has bar/345 as parent

final call `drive_list('567' in parents)`

R/drive_ls.R
jennybc: Doesn't it still boil down to this: Does folder i have folder i - 1 as parent? You can get all of that information at once, I think. There's no need for a while.

LucyMcGowan:
okay so I *think* I have something working. 
* created a new function [`drive_list()`](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_list.R)
* formatted as suggested [here](https://github.com/tidyverse/googledrive/pull/13#issuecomment-301207813)
* removed all the helper functions, as suggested and replaced with calls to `drive_list()` (except when figuring out the [root](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_list.R#L49), but this will probably change in next iteration)
* it basically [loops through](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_list.R#L49-L58), starting at the top folder, & sees if it is in the proper folder above. There is likely a more elegant way to do this, but I haven't thought of it yet 😬

R/drive_ls.R
LucyMcGowan: yes - I think I've got something similar to that in my [newest iteration](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_list.R#L49-L58) - one tricky part is that you may have x folders named `foo` and y named `bar` etc. so it isn't always checking if it's in i-1, it could be that you have something like
```
name  location
foo      root
foo      bar
bar      foo
baz     bar
```
but I *think* I've solved that by [only setting the `parent_id` if it is correct](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_list.R#L53-L56). I would like to refactor this so it isn't a for loop (which I'm sure is very doable) but I am a bit sleepy. Thank you for walking through this with me I feel like I've got you really getting into the weeds on this one 🌾

LucyMcGowan:
to take it for a little spin to make sure we're seeing the same results:
```
devtools::install_github("tidyverse/googledrive",ref = "drive_list")

#assuming you have the following file structure
# foo
# Lbar
# Lbaz
# Lyo my_file

#this should return a folder "bar"
drive_list(path = "foo")

#this should return a folder "baz"
drive_list(path = "foo/bar")

#this should return a folder "yo" and a file "my_file"
drive_list(path = "foo/bar/baz")

#let's say I just wanted to get the file, I could pass to the pattern
drive_list(path = "foo/bar/baz", pattern = "my_file")

#or I could pass to `q` to only get the documents (this will be a feature soon)
drive_list(path = "foo/bar/baz", q = "mimeType='application/vnd.google-apps.document'")
```

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
+#' List files on Google Drive
+#'
+#' @param path character vector, path name(s) where the Google drive files are
```
jennybc: This function won't be vectorized over `path`, so I would delete the `(s)` to avoid any confusion. If someone needs to do this for several paths, they'll need to map.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+#'   that you would like to list. Defaults to the "My Drive" directory.
+#' @param pattern character vector, regular expression(s) of title(s) of
```
jennybc: I know I am probably the source of this, because I accepted multiple regexes in `gs_ls()`. In hindsight, I think it's preferable to play more dumb and only accept **a** regex. If people want to catenate multiple regexes, they can do this themselves. Plus, in this function, the whole power of `q` is just one argument away!

I would be like `pattern` in `list.files()` and accept only one regex.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+#'   documents to output in a tibble. If it is `NULL` (default), information
+#'   about all documents in drive will be output in a tibble.
```
jennybc: I would eliminate the last sentence, because many other arguments might be filtering the results.

I'm going to make a high-level comment on this elsewhere but given the power of `q`, what is the point of `pattern`? All other things being equal, it makes sense to implement all filters via `q` to make your package snappier. But I think `pattern` is required to filter names on a regular expression, right? That's why we implement this *post hoc* on the R side.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+#' }
+#'
+#' @export
+drive_list <- function(path = NULL, pattern = NULL, ..., verbose = TRUE){
+
+  folder <- NULL
+  if (!is.null(path)){
+    folder_pattern <- paste0("^", gsub("/", "$|^", path), "$")
+    folder_nms <- unlist(strsplit(path, "/"))
+    folder_order <- tibble::tibble(name = folder_nms,
```
jennybc: tidyverse style would be like so:

``` r
folder_order <- tibble::tibble(
  name = folder_nms,
  dir = 1:length(folder_nms)
)
```

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+                                   dir = 1:length(folder_nms))
```
jennybc: `seq_along(folder_nms)` is safer. In general, prefer `seq_along()` and `seq_len()` in these situations.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+    folder_tbl <- drive_list(pattern = folder_pattern,
+                             fields = paste0("files/parents,files/name,files/mimeType,files/id"),
```
jennybc: Why the surrounding `paste0()`?

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+                             q = "mimeType='application/vnd.google-apps.folder'")
+
+    if (!all(folder_nms %in% folder_tbl$name)){
+      spf("We could not find the file path '%s' on your Google Drive", path)
```
jennybc: Technically, this should probably glue the path components up to and including the first one that is not found. So if we're looking for `foo/bar/baz` and `foo` exists but has no folder child named `bar`, this error will complain about `foo/bar` instead of just `bar`.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    # merge in to get the correct directory order & sort, oh how I miss the %>% :(
```
jennybc: You should import and re-export `%>%` from purrr, which you already import. Then I think it's fine to use internally, in moderation. Put lines [like these](https://github.com/tidyverse/googlesheets2/blob/8caf705a92b7c0638783ce478eb0449a3a92c1de/R/utils.R#L1-L9) in your `utils.R`.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+    folder_tbl <- tibble::as_data_frame(merge(folder_tbl, folder_order, by = "name"))
```
jennybc: I think `tibble::as_tibble()` is the thing now?

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+      if (keep){
+        parent_id <- subfolder$id
+      }
+      keep_folders[i] <- keep
+    }
+
+    folder_tbl_keep <- folder_tbl_sort[keep_folders, ]
+    leafmost <- nrow(folder_tbl_keep)
+    if (leafmost != length(folder_nms)){
+      spf("Uh oh, something went wrong. We could not find the path '%s' on your Google Drive", path)
```
jennybc: What's the scenario in which this happens? Let's be more specific for our future selves and in the message.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+#' @examples
+#' \dontrun{
+#' ## list user's Google Sheets
+#' drive_list(q = "mimeType='application/vnd.google-apps.spreadsheet'")
+#' }
+#'
+#' @export
+drive_list <- function(path = NULL, pattern = NULL, ..., verbose = TRUE){
+
+  folder <- NULL
```
jennybc: Looping back after reading all the path-processing stuff ... ok so now I do advocate moving all this logic into a helper function. It's the internal helper that takes a path and returns id for leafmost directory. Input is `path` (and possibly `...`? IDK). Output is something that needs to be incorporated into `q`.

R/drive_list.R
```diff
@@ -0,0 +1,162 @@
...
+    if (leafmost != length(folder_nms)){
+      spf("Uh oh, something went wrong. We could not find the path '%s' on your Google Drive", path)
+    }
+
+    folder <- folder_tbl_keep[leafmost, "id"]
+  }
+  x <- list(...)
+
+  #add default fields if null
+  if(is.null(x$fields)){
```
jennybc: Style: `if (is.null(x$fields)) {`

But more importantly ... you might want to store this list elsewhere to keep `drive_list()` more readable. It can even be somewhere in this file, but at the bottom, where you can easily consult it but can also ignore it. It's a good thing to store in an internal environment.

R/drive_list.R
LucyMcGowan: I agree with removing the last sentence. I *think* how we are going about it with `pattern` now makes sense in order to implement regular expressions (google drive lets one search using [contains, =, !=](https://developers.google.com/drive/v3/web/search-parameters), which probably serves most people's uses, but I could imagine some cases where our regex capability is quite useful.

R/drive_list.R
LucyMcGowan: 🙈 I was pasting them together with `,` in between previously and forgot to remove that

R/drive_list.R
LucyMcGowan: This could happen in the case that you have something like `path = "foo/bar/yo"` where all of those folders exist somewhere on your drive, but not in the nested way as suggested (for example the real path would be `foo/bar/baz/yo`). In this case, no errors would be thrown up until this point because all of the folders exist, but when we try to find folders within the folders, only `foo` and `bar` would come out of this function since `yo` is not in `bar`. I've [updated to be more informative](https://github.com/tidyverse/googledrive/blob/drive_list/R/drive_list.R#L125-L128).

R/drive_list.R
```diff
@@ -0,0 +1,206 @@
...
+
+  #add folder if not null
+  if (!is.null(folder)){
+    q <- paste0("'", folder, "'", " in parents")
+    if (!is.null(x$q)){
+      x$q <- paste(x$q, "and", q)
+    } else {
+      x$q <- q
+    }
+  }
```
jennybc: This makes me think you will want to make some tools for validating and building up `q`. But let's not worry about this in this PR, which is already a beast 👹!

R/drive_list.R
```diff
@@ -0,0 +1,206 @@
...
+  req_tbl <- tibble::tibble(
+    name = purrr::map_chr(proc_res$files, "name"),
+    type = sub(".*\\.", "", purrr::map_chr(proc_res$files, "mimeType")),
+    folder_id = purrr::map_chr(
+      purrr::map(
+        proc_res$files,
+        "parents",
+        .null = NA
+      ),
+      1),
```
jennybc: I don't understand the rationale for renaming `parents` to `folder_id`.

R/drive_list.R
```diff
@@ -0,0 +1,206 @@
...
+    id = purrr::map_chr(proc_res$files, "id"),
+    gfile = proc_res$files)
+
+  if (is.null(pattern)){
+    return(req_tbl)
+  } else{
+    if (!(inherits(pattern, "character") & length(pattern) == 1)){
+      stop("Please update `pattern` to be a character string.")
+    }
```
jennybc: This should be done at the vveerryy beginning, before nothing with an API calls, for example.

R/drive_list.R
```diff
@@ -0,0 +1,206 @@
...
+  folders$folder_id <- NULL
+
+  if (!all(path_pieces$name %in% folders$name)){
+    spf("We could not find the file path '%s' on your Google Drive", path)
+  }
+  ## guarantee: we have found at least one folder with correct name for
+  ## each piece of path
+
+  folders <- tibble::as_tibble(merge(folders, path_pieces, by = "name"))
+  folder <- folders$id[folders$depth == d]
```
jennybc: This is the candidate return value.

R/drive_list.R
jennybc: I have taken the liberty of putting this in a package-wide environment `.drive` that we will use to hold info on the API. I've moved the URLs  into `.drive` and out of `.state`, which we can reserve for auth stuff. The `.drive` environment is also where we'll eventually put a tibble of mimeTypes and human-friendly names, for example.

jennybc:
@LucyMcGowan OK I've made some more changes and put some bugs/problems into comments. So please have a look at my more recent review comments and my commits. We're getting there!

R/drive_list.R
```diff
@@ -0,0 +1,191 @@
...
+  if (!all(path_pieces %in% folders$name)){
+    spf("We could not find the file path '%s' on your Google Drive", path)
+  }
+  ## guarantee: we have found at least one folder with correct name for
+  ## each piece of path
+
+  folders$depth <- match(folders$name, path_pieces)
+  folders <- folders[order(folders$depth), ]
+  folder <- folders$id[folders$depth == d]
+  if (length(folder) != 1) {
```
LucyMcGowan: I think this would eliminate the case where you may have `foo/bar/baz` and also `foo/yo/baz`, which should be allowed, right?

R/drive_list.R
```diff
@@ -102,10 +103,13 @@ get_leafmost_id <- function(path) {
...
-  if (length(folder) != 1) {
-    spf("'%s' does not uniquely define a single path", path)
-  }
+  # if (length(folder) != 1) {
+  #   spf("'%s' does not uniquely define a single path", path)
+  # }
   ## guarantee: there is exactly one folder at depth d
+  ## I actually think that is too conservative, because you could have
+  ## "foo/bar/baz" and "foo/yo/baz" both in your drive, and that would
+  ## be perfectly kosher, but we wouldn't allow it here with this error
```
jennybc: Good point! When I resolved our merge conflicts, I removed all of this.

R/drive_list.R
```diff
@@ -124,12 +128,32 @@ get_leafmost_id <- function(path) {
...
+
+  # if there are multiple in depth d & it isn't the root
+  if (length(folder) > 1) {
+    leafmost_parent <- folders[folders$depth == d - 1, ]
+    child_is_leafmost <- purrr::map_lgl(
+      folders$parents,
+      ~ .x == leafmost_parent$id
+    )
+    folder <- folders$id[child_is_leafmost]
+  }
```
jennybc: I'm a bit foggy on the scenario we are guarding against here, but I don't disagree. Would be good to write it down specifically and incorporate into a test. All of this will be much easier to code and discuss when there are tests with concrete scenarios for directory structure and desired result.

jennybc:
So close!

It feels like all that's really missing are tests.

We have a chicken and egg problem:

The functions for creating folders and files aren't the current focus and may be very tentative. But we can't move on until we merge this.

However, ideally the tests would create the complicated structures to test against, so that they run for anyone.

In the name of pragmatism, perhaps you should record the various scenarios that have come up so far. Describe them in enough detail that you or I can make sure our Drive has the necessary stuff. Then update the demo to explore all of them. Then at least we can each test informally by walking through the demo and merge this.

Then we can open narrower issues for next steps.

R/drive_list.R
jennybc: Correct! That one was not on my radar. Removed this stuff in the merge.

LucyMcGowan:
I've added a few [tests](https://github.com/tidyverse/googledrive/blob/drive_list/tests/testthat/test-drive_list.R) currently toggled with `internet = TRUE` that capture the few cases we've been discussing. I also ended up adding a function `drive_mkdir()`, since I was doing this a lot in the tests (I was hoping not to add much more to this PR, but it was getting tiresome to create directories without it 🙈)

R/drive_list.R
```diff
@@ -80,6 +80,18 @@ drive_list <- function(path = NULL, pattern = NULL, ..., verbose = TRUE){
...
+
+  if (all(path_pieces == "~")) {
+    return(root_id)
+  }
+
+  ## remove ~ if it was added to the beginning of the path (like ~/foo/bar/baz instead of foo/bar/baz)
+  if (path_pieces[1] == "~") {
+    path_pieces <- path_pieces[-1]
+  }
+
```
jennybc: I suspect it would make more sense to deal with (most of?) this in the caller, i.e. checking and preparing `path`. This check + prep is a good candidate for a helper function, because I bet it comes up elsewhere.

R/drive_mkdir.R
```diff
@@ -0,0 +1,83 @@
+#' Make a Google Drive Directory
+#'
+#' @param dir character, name of the folder you would like to create
+#' @param path character, path where you would like the folder on your Google Drive
+#'
+#' @return object of class `gfile` and `list` that contains uploaded folder's
+#'   information
+#' @export
+drive_mkdir <- function(dir = NULL, path = NULL) {
```
jennybc: To consider: should this be named -- or at least aliased -- to `drive_dir_create()` because `dir.create()`?

LucyMcGowan: oh probably! also, that is much more explicit for non-unix friends, I like it

PR #20, opened 2017-05-16
bake api_url into build fx, fixes #16

### LINKED ISSUE #16: Bake in the base URL for Drive API
I just changed where the base URL for Drive API is stored and it caused changes in most of the `.R` files. That doesn't seem desirable.

I don't think this package makes any calls to URLs that are not under the main Drive base URL. If I've got that right, then the base URL should get baked into the function that builds requests.

A slight variant of this is the files URL. That should probably NOT be stored anywhere, but, instead, get built into requests by specifying `path = "files"` (in the API sense, not the `drive_list()` sense!) 

Added later: I think, in fact the `api_url`, is ready and waiting to be used here. I see it in every built request.

LucyMcGowan:


R/aa_utils.R
```diff
@@ -10,8 +10,6 @@ NULL
 
 # environment to hold data about the Drive API
 .drive <- new.env(parent = emptyenv())
-.drive$base_url_files_v3 <- "https://www.googleapis.com/drive/v3/files"
-.drive$base_url <- "https://www.googleapis.com"
```
jennybc: I think you might still want to store the URL here. Although atm I can't think of a specific reason why!

R/drive_mv.R
```diff
@@ -37,20 +37,16 @@ build_drive_mv <-
...
-    build_request(endpoint = url,
-                  token = token,
-                  method = "PATCH")
+    build_request(endpoint = paste0(
+      file$id,
+      "?addParents=",
+      folder$id,
+      "&removeParents=",
+      file$kitchen_sink$parents[[1]]
+    ),
```
jennybc: This all feels like info that should be provided via the query parameters vs. `paste0()` on the base url.

R/drive_publish.R
```diff
@@ -53,9 +53,9 @@ build_drive_publish <-
     id <- file$id
 
     rev_id <- file$publish$revision
-    url <- file.path(.drive$base_url_files_v3, id, "revisions", rev_id)
+    endpoint <- file.path(id, "revisions", rev_id)
     build_request(
-      endpoint = url,
+      endpoint = endpoint,
```
jennybc: Again, this feels like some concatenation that can be handled via normal httr URL handling, if the info is stored in the expected place.

R/drive_publish.R
```diff
@@ -110,9 +110,9 @@ build_drive_check_publish1 <-
 
     id <- file$id
 
-    url <- file.path(.drive$base_url_files_v3, id, "revisions")
+    endpoint <- file.path(id, "revisions")
 
-    build_request(endpoint = url,
+    build_request(endpoint = endpoint,
```
jennybc: Ditto?

R/drive_mv.R
LucyMcGowan: this is kind of a problem with my internal functions, but right now in order for `params` to get passed to the `query`, it must use the `GET` method, otherwise the `params` pass to the body.

R/workhorse.R
```diff
@@ -1,8 +1,9 @@
-build_request <- function(endpoint = NULL,
+build_request <- function(endpoint = NULL ,
+                          path = "drive/v3/files",
                           params = list(),
                           token = NULL,
                           send_headers = NULL,
-                          api_url = NULL,
+                          api_url = "https://www.googleapis.com",
```
jennybc: This could still be `.drive$base_url <- "https://www.googleapis.com"` but, as I said, I can't exactly say why I would store it in `.drive`

LucyMcGowan:
I set the default `path` to the files and then was using the `endpoint` to add to it - do you think I should have another word (leave `path` as the base, and only change it if we don't want to go to `files`, and find a different word for `endpoint` when I'm tacking things onto the end of `path`?

R/workhorse.R
```diff
@@ -44,21 +42,25 @@ set_query <- function(x){
...
-  if (grepl("^http", x$endpoint)){
-    x$url <- x$endpoint
+  if (!is.null(x$endpoint)) {
+    x$url <- file.path(x$api_url,
+                       x$path,
+                       x$endpoint)
   } else{
-    x$url <- file.path(x$api_url, sub(".", "", x$endpoint))
+    x$url <- file.path(x$api_url,
+                       x$path)
```
jennybc: I think you should be using `httr::modify_url()` for this, not `file.path()`.

More like so:
https://github.com/tidyverse/googlesheets2/blob/8caf705a92b7c0638783ce478eb0449a3a92c1de/R/gs_build_request.R#L84-L102

Let me know if I'm missing some big difference between these two APIs.

R/drive_mv.R
jennybc: That sounds like something worth fixing at the source, i.e worth refactoring the request builder. It's up to you if you want to do that here or ... do it this way for now and simultaneously open an issue for the other.

jennybc:
I think `path` should be the string that needs to be tacked onto the URL. In Google Sheets, this often requires substituting some user-provided "path" parameters, such as the spreadsheet id. I think that's pretty similar to your situation. So one of the main jobs of my request builder is to make path vs query parameters from the user-provided parameters. I suspect you should do same?

LucyMcGowan:
@jennybc I'm a little bit excited about some of the updates I've been working on today 🎉 . I ended up pulling all of the docs into a tibble to reference from within the `build_request()` so that it knows if it should pass the parameter to the query or body (also this is a good check for the parameters passed). Let me know what you think!

R/drive_mkdir.R
```diff
@@ -16,13 +16,13 @@ drive_mkdir <- function(dir = NULL, path = NULL, verbose = TRUE) {
   }
 
   request <- build_request(
+    method = "create",
```
jennybc: It's unexpected to see anything other than an HTTP verb here. What's up?

R/drive_upload.R
```diff
@@ -176,14 +176,17 @@ build_drive_upload <- function(input = NULL,
     id <- proc_res$id
   }
 
-  list(
-    method = "PATCH",
-    endpoint =  paste0(id, "?uploadType=media"),
-    path = "upload/drive/v3/files",
+  build_request(
+    path = "https://www.googleapis.com/upload/drive/v3/files/{fileId}",
```
jennybc: I'm still getting my bearings, so humor me. Overall the changes look good. But here I'd expect `https://www.googleapis.com` to be coming from a built-in package default and `path` would hold just `/upload/drive/v3/files/{fileId}`.

jennybc:
I've only taken a quick peek but it seems like a great step forward. 😄

I think I could use some vocabulary orientation. When I see `method` I think GET, POST, PATCH, etc. but obviously that's not how you're using it. And I know that the Google API docs also don't necessarily use it that way. Can you could help me out with some high-level orientation?

LucyMcGowan:
I used `method` like the [Google API docs] (https://developers.google.com/drive/v3/reference/) do, since I scraped the tables from there, but maybe I should change that if most people thing of the httr verbs as methods. It is all set up as the [API docs](https://developers.google.com/drive/v3/reference/), so `resource` refers to the resources laid out (in the nav bar on the left of the docs) and `methods` refers to which method you'd like to perform as defined by the initial [API doc tables](https://developers.google.com/drive/v3/reference/).

R/drive_upload.R
LucyMcGowan: oh agreed - I'll change that!

LucyMcGowan:
This may help with re-orientation - here is a list of `method`s & `resource`s (I am totally open to changing the wording for `method` though).
|resource    |method            |
|:-----------|:-----------------|
|about       |get               |
|changes     |getStartPageToken |
|changes     |list              |
|changes     |watch             |
|channels    |stop              |
|comments    |create            |
|comments    |delete            |
|comments    |get               |
|comments    |list              |
|comments    |update            |
|files       |copy              |
|files       |create            |
|files       |delete            |
|files       |emptyTrash        |
|files       |export            |
|files       |generateIds       |
|files       |get               |
|files       |list              |
|files       |update            |
|files       |watch             |
|permissions |create            |
|permissions |delete            |
|permissions |get               |
|permissions |list              |
|permissions |update            |
|replies     |create            |
|replies     |delete            |
|replies     |get               |
|replies     |list              |
|replies     |update            |
|revisions   |delete            |
|revisions   |get               |
|revisions   |list              |
|revisions   |update            |
|teamdrives  |create            |
|teamdrives  |delete            |
|teamdrives  |get               |
|teamdrives  |list              |
|teamdrives  |update            |

jennybc:
Thanks, this will help. I can't digest it right now, but I went through similar re-re-naming cycle with googlesheets2 already. I'll read the above, see what I settled on over there, and come back. We should try to be consistent, so it's easier to work on these two packages in tandem.

R/workhorse.R
```diff
@@ -1,18 +1,16 @@
 build_request <- function(path = NULL,
                           params = list(),
-                          method = "list",
-                          resource = "files",
+                          endpoint = "drive.files.list",
                           token = NULL,
```
jennybc: If the default value of `token` were `drive_token()`, that would remove a lot of repetition, yes?

LucyMcGowan: yes! 🙈

R/workhorse.R
```diff
@@ -34,9 +31,8 @@ set_path <- function(x) {
   if (!is.null(x$path)) {
     return(x)
   }
-  ## find the path that matches the given method and resource
-  x$path <- unique(.drive$params$path[.drive$params$method == x$method &
-                                        .drive$params$resource == x$resource])
+  ## find the path that matches the given endpoint
+  x$path <- unique(.drive$params$path[.drive$params$endpoint == x$endpoint])
   return(x)
```
jennybc: I'm not sure of the exact structure of `.drive$params` but could you take an approach like I've done here:

https://github.com/tidyverse/googlesheets2/blob/8caf705a92b7c0638783ce478eb0449a3a92c1de/R/gs_build_request.R#L47

where you store the row or component corresponding to this endpoint as, say, `ept`. Then get to make much simpler calls to extract bits? The `unique()` catches my eye. Are we actually at risk of having things ... of length greater than 1?

tests/testthat/.Rapp.history
```diff
@@ -0,0 +1 @@
+load("/Users/lucymcgowan/Documents/wonderland/googledrive/tests/testthat/testing-token.rds")
```
jennybc: This should be removed.

R/workhorse.R
LucyMcGowan: right now `.drive.params` is a tibble that has a separate row with every parameter allowed for each endpoint, so it is just grabbing the endpoints is >1 (but they should all have the same value). I could save it in 2 separate tibbles (one for the paths, which are unique to each endpoint and one for the parameters?)

tests/testthat/test-build_request.R
```diff
@@ -0,0 +1,95 @@
...
+                     glue::glue_data(
+                       tibble::enframe(params_wrong),
+                       "{name}: {value}"
+                     )
+                   ),
+                   collapse = "\n")
+  )
+
+
+})
```
jennybc: These look good. My only advice would be, again, to split into smaller tests with pretty tight goals, if you can.

tests/testthat/test-drive_publish.R
```diff
@@ -0,0 +1,32 @@
...
+  drive_chickwts <- drive_publish(drive_chickwts, publish = FALSE)
+
+  ## now this sould be false
+  expect_false(drive_chickwts$publish$published)
+
+  ## clean up
+  drive_delete(drive_chickwts)
+  rm <- file.remove("chickwts.txt")
+})
+
```
jennybc: Looks good.

R/workhorse.R
jennybc: Hmm.... I store my API info as a list at this point, with one component per endpoint. Then it has components for `path`, `method` (in the GET, POST, etc.) sense, and `params`, which is a list.

I tend to think this is a good idea, because ... well, that's where I eventually came out 😄 after going through my own tibble phase. But let's not let this block merging this PR. You could open another issue for "refactor storage of API data" or some such, to put this on the radar.

jennybc:
This PR probably needs a better title, because it acquired a new, more serious goal as it matured.

R/drive_file.R
```diff
@@ -7,68 +7,17 @@
 #' @export
 #'
 drive_file <- function(id = NULL, ...) {
-  request <- build_drive_file(id = id, ...)
+  request <- build_request(
+    endpoint = "drive.files.get",
+    token = drive_token(),
+    params = list(...,
```
jennybc: If the user specified `fileId = "abcdef"` in `...`, would `build_request()` catch it? Are you checking for params that appear more than once?

R/drive_share.R
```diff
@@ -49,48 +42,38 @@ drive_share <- function(file = NULL,
...
                               type = NULL,
-                              email = NULL,
-                              message = NULL,
                               ...,
                               token = drive_token()) {
   if (!inherits(file, "gfile")) {
     spf("Input must be a `gfile`. See `drive_file()`")
   }
 
+  if (is.null(role) | is.null(type)) {
```
jennybc: Suspect this should be `||`.

R/drive_upload.R
```diff
@@ -179,17 +176,17 @@ build_drive_upload <- function(input = NULL,
...
 
-  url <- file.path(.drive$base_url,
-                   "upload/drive/v3/files",
-                   paste0(id, "?uploadType=media"))
-
-  list(
-    method = "PATCH",
-    url = url,
+  build_request(
+    path = "/upload/drive/v3/files/{fileId}",
```
jennybc: I've gotten the impression that `path` is now looked up based on endpoint. Is something different here on purpose?

R/workhorse.R
```diff
@@ -1,78 +1,145 @@
-build_request <- function(endpoint = NULL,
+build_request <- function(path = NULL,
```
jennybc: Is `path` still needed? Is it here in case someone want to use this for an endpoint you have somehow not put into your internal data about the API?

R/drive_upload.R
LucyMcGowan: there are 2 endpoints that have 2 different paths depending on what you are trying to do (both have to do with uploads). If you are just updating the metadata, you use the path "/drive/v3/files/{fileId}" (that is what we are doing most of the time), but if you are updating the *actual* file (like upload new content, as we are here), you have to change the path to "/*upload*/drive/v3/files/{fileId}"

R/workhorse.R
LucyMcGowan: we need it now for the 2 endpoints that don't have unique paths (`.drive.files.create` and `.drive.files.update`) we could break from the Google API docs and make up our own endpoints to make them unique if you think that'd be better

jennybc: You might want to consider a two-layer request builder, like I made here:

https://github.com/tidyverse/googlesheets2/blob/8caf705a92b7c0638783ce478eb0449a3a92c1de/R/gs_build_request.R#L47

The inner layer, `gs_build_request()`, is very dumb and mechanical. But would work if someone really knew what they were doing and only wanted help setting base URL, path param expansion, and managing auth.

The outer layer, `gs_generate_request()`, is smarter and what I plan to use! It needs to be given a known endpoint and then it looks up everything we now, checks that input looks good, and massages it into shape before calling `gs_build_request()`.

If you're game, you can make that a separate PR. Let's get this one merged.

R/drive_file.R
LucyMcGowan: It *sort* of ends up working out (but I can add more structure around this) - if a user passes a `fileId` and the endpoint is not one that takes a fileId, they will receive a message that the fileId parameter wasn't use. If the user supplies a fileId and a fileId *is* expected, the fileId I've told it to use takes precedence, so their fileId essentially does nothing (but I don't tell them that, so maybe I should) (I've added a test for this in my next push)

jennybc: OK don't hold this PR up. I've tried to sell you on my request-building approach from googlesheets2. If you buy what I'm selling, you can take that logic from there as well (the checking for repetition). In Sheets, some repetition is actually allowed but AFAICT that's not true for Drive.

R/drive_upload.R
jennybc: If I'm following you (I've also read the comment below), YES we should create distinct endpoints. So that identification by endpoint tells you exactly what `path` should be, exactly what the http verb is, what params are allowed, etc.

jennybc: We're talking about this, yes?

https://developers.google.com/drive/v3/reference/files/update

Maybe `drive.files.update.meta` vs `drive.files.update.media`?

R/drive_file.R
LucyMcGowan: definitely buying it! I'm adding in the repetition & unique endpoint (re the comment 👇 ) and I'll merge! thank you!

R/drive_upload.R
LucyMcGowan: yes! I've update to that in the next push - great idea 💡

Issue #12, opened 2017-05-12
README.Rmd


jennybc:
Also not urgent, but good to initiate once the UI starts to settle down. Below is boilerplate content, so use your judgment.

### `README.Rmd`

  * [x] Switch from `.md` to `.Rmd`

  * [x] Install instructions:

````r
```{r, eval = FALSE}
# The easiest way to get readr is to install the whole tidyverse:
install.packages("tidyverse")

# Alternatively, install just readr:
 install.packages("readr")

# Or the the development version from GitHub:
# install.packages("devtools")
devtools::install_github("tidyverse/readr")
```
````

  * [ ] Overview: 1-2 paragraphs including brief description/goals. Add h2, and set:

   ```yaml
   home:
     strip_header: true
   ```

  * [ ]  Link to appropriate R4DS chapter:

   ```md
   If you are new to readr, the best place to start is the
   [data import chapter](http://r4ds.had.co.nz/data-import.html) in R
   for data science.
   ```

  * [ ] Add link to `_pkgdown.yaml`

   ```yaml
     links:
     - text: Learn more
       href: http://r4ds.had.co.nz/data-import.html
   ```

  * [x]  Usage section shows a few examples.

#

**Comment by LucyMcGowan** [comment]
@jennybc does this go in the `_pkgdown.yaml`? I am not sure where to stick it for the README.
```
home:
  strip_header: true
```

**Comment by jennybc** [comment]
I'm not even sure of the context, yet I note that I have this in the [readxl `_pkgdown.yaml`](https://github.com/tidyverse/readxl/blob/cb2b98144f2e7bb7374a08435c1286c46baeeefd/_pkgdown.yml).

What's the connection / need re: the README?

**Comment by LucyMcGowan** [comment]
I'm not sure, it's from the third checklist item above:
> Overview: 1-2 paragraphs including brief description/goals. Add h2, and set:
```
home:
  strip_header: true
```

**Comment by jennybc** [comment]
It must be something in the pkgdown setup that is relevant to pkgdown's treatment of README, which is basically the basis for index.html by default. So I think you can disregard until you're using  pkgdown.

**Comment by jennybc** [comment]
In #34 I also added an explicit token load to the setup chunk for `README.Rmd`. This way we are both using the dev account. By way of explanation, you need to capture a token for that account that is *noncaching*, i.e. it doesn't write itself to `.httr-oauth` every time it refreshed. You will need to this:

```
[...]

PR #21, opened 2017-05-16
update README (#12)

LucyMcGowan:


README.Rmd
```diff
@@ -8,18 +8,120 @@ output: github_document
...
+[![Build Status](https://travis-ci.org/tidyverse/googledrive.svg?branch=master)](https://travis-ci.org/tidyverse/googledrive)[![AppVeyor Build Status](https://ci.appveyor.com/api/projects/status/github/tidyverse/googledrive?branch=master&svg=true)](https://ci.appveyor.com/project/tidyverse/googledrive)[![Coverage Status](https://img.shields.io/codecov/c/github/tidyverse/googledrive/master.svg)](https://codecov.io/github/tidyverse/googledrive?branch=master)
 
 `r emo::ji("construction")` WARNING: this is very much under construction `r emo::ji("construction")`
 
-## installation
+## Overview
+
+`googledrive` interfaces with Google Drive from R, allowing users to seamlessly manage files on Google Drive from the comfort of their console `r emo::ji("house")`. 
+
```
jennybc: Technically, you'll need to suggest emo if you do this and it may not be worth it?

README.Rmd
```diff
@@ -8,18 +8,120 @@ output: github_document
...
+* All functions begin with the prefix `drive_`
+* Functions and parameters attempt to mimic local file navigating conventions in `R`, such as `list.files`.
+
+### Quick demo
+
+Here's how to list the most recently modified 100 files on your drive. This will kick off your authentication, so you will be sent to your browser to authorize your Google Drive access. The functions here are designed to be pipeable, using `%>%`, however they can also be implemented without.
+
+```{r, eval = FALSE}
+drive_list()
+```
```
jennybc: It would be nice to actually run these chunks, so output appears. If you don't want to do that now, then put it on the to do list.

README.Rmd
```diff
@@ -8,18 +8,120 @@ output: github_document
...
+
+Notice the access here says "Shared with specific people". To update the access, we need to change the sharing permissions. Let's say I want anyone with the link to be able to view my dataset.
+
+```{r}
+drive_chickwts <- drive_chickwts %>%
+  drive_share(role = "reader", type = "anyone")
+```
+
+*Notice when I run any of the `googledrive` functions, I am assigning them. This is good practice, since each of these functions is inherintely changing our file on Google Drive, so we want to keep track of that with our `R` object.*
+
```
jennybc: spelling: inherently

I'd reword more along these lines:

> We always assign the return value of googledrive functions back into an R object. This object is of type `gfile`, which holds up-to-date metadata on the associate Drive file. By constantly re-assigning the value, we keep it up-to-date, facilitating all downstream operations.

Tidyverse convention is to surround functions with backticks but not the names of packages or R itself.

README.Rmd
```diff
@@ -8,18 +8,120 @@ output: github_document
...
+We can then extract a share link.
+
+```{r}
+drive_chickwts %>%
+  drive_share_link
```
jennybc: Convention is to include the trailing parens, e.g. `drive_share_link()`

README.Rmd
LucyMcGowan: Do you think it'd be strange to see a bunch of my files? I suppose it doesn't matter

README.Rmd
LucyMcGowan: 😩 agreed not worth it

README.Rmd
jennybc: I have sometime just let this happen with googleshets, but yes I know what you mean. This is another reason for us to start using a dedicated Google account for development (the one I mentioned today).

The README, tests, and vignettes should all explicitly load a token associated with your desired "development" account. Then you can just let chunks like this run.

Let me know if you want those credentials. We'll just have to be careful to not delete any of the files that googlesheets tests against!

It's possible we should create a second account for googledrive development?

README.Rmd
jennybc: 😭

README.Rmd
jennybc: Maybe you don't need to repeat "up-to-date" so many times. Oops! But you get the idea.

README.Rmd
jennybc: You are welcome to leave this un-eval'd and open issue about the use of a common development Google account and associated token. In fact, you should open that issue no matter what, if we don't have it already.

LucyMcGowan: In the one I just pushed I think I fixed that?

README.Rmd
```diff
@@ -41,25 +41,25 @@ library("googledrive")
...
 
 Here's how to list the most recently modified 100 files on your drive. This will kick off your authentication, so you will be sent to your browser to authorize your Google Drive access. The functions here are designed to be pipeable, using `%>%`, however they can also be implemented without.
 
-```{r, eval = FALSE}
+```{r}
 drive_list()
 ```
 
-You can narrow the query by specifying a `path` and/or `pattern` you'd like to search within. For example, to search within the folder `foo` for a file named `bar` you could run the following.
+You can narrow the query by specifying a `path` and/or `pattern` you'd like to search within. For example, to search within the folder `foobar` for a file named `baz` you could run the following.
```
jennybc: This makes me wonder, as a user, why can't I just provide `foobar/baz`? What would be the difference? 🤔 I suspect answering this has implications for `drive_list()` docs and/or logic. Maybe put this in the existing issue thread for `drive_list()` interface? It has nothing to do with the README.

jennybc:
I made more comments but LGTM. Merge away!

README.Rmd
LucyMcGowan: @jennybc 🤔 that is a good point, the way we have the logic now we expect the `path` parameter to always be a folder. In the broad sense, the pattern would usually be something like you'd want to find all the files with `baz` in them in a specific folder? But I can see a user wanting to specify a specific file and being sad that this doesn't let you do that with just one parameter. But then also if you are trying to find a specific file is that the best use of a "list"? (as you can see I've talked myself right around in a circle here). If keeping with `list.files` I don't believe it lets you pass a full path to the first parameter, it expects a directory and lists everything inside it.

LucyMcGowan: but maybe this is a good indication that I should reword the example for more of that functionality (like add in 2 things with the characters `baz` in the name into the `foobar` folder)

PR #22, opened 2017-05-16
add check for content fixes #14

### LINKED ISSUE #14: check content
* note to myself to remember to do this

LucyMcGowan:
* I added a check for content as `application/json; charset=UTF-8` 
* I think this is true for all of my output thus far, but I've allowed for input of `expected` (default = "application/json; charset=UTF-8"), `encoding` (default = "UTF-8"),  and `as` (default = "parsed")


R/workhorse.R
```diff
@@ -27,57 +28,79 @@ build_request <- function(endpoint = NULL,
 ## right now not setting endpoint, if you want to include parameters, must be in a named list
 ## may introduce the :parameter notation in endpoint later :)
 
-set_query <- function(x){
-  if (length(x$params) == 0L) return(x)
+set_query <- function(x) {
+  if (length(x$params) == 0L)
+    return(x)
```
jennybc: I'm a believer in always using brackets, just to be safe.

R/workhorse.R
```diff
@@ -27,57 +28,79 @@ build_request <- function(endpoint = NULL,
...
   x
 }
 
 ## not setting headers yet
 
-set_url <- function(x){
-
-  if (grepl("^http", x$endpoint)){
+set_url <- function(x) {
+  if (grepl("^http", x$endpoint)) {
```
jennybc: This smells like something that will go away once this is designed to align perfectly against `httr::modify_url()`.

R/workhorse.R
```diff
@@ -27,57 +28,79 @@ build_request <- function(endpoint = NULL,
...
+        paste0(
+          "Expected content-type:\n%s",
+          "\n",
+          "Actual content-type:\n%s"
+        ),
+        expected,
+        actual
+      )
+    }
+    httr::content(res, as = as, encoding = encoding)
```
jennybc: I think this is *safe*: you're checking that content declares itself as UTF-8 JSON then you let `httr::content()` auto-parse. I think it is slightly unconventional? For example, if you read the docs for `httr::content()`, `as = "parsed"` is discouraged in a package. But, as I said, I think you are doing the necessary checks.

R/workhorse.R
```diff
@@ -27,57 +28,79 @@ build_request <- function(endpoint = NULL,
...
   x$query <- x$params
   x$params <- NULL
   x
 }
 
-set_body <- function(x){
-  if (length(x$params) == 0L) return(x)
+set_body <- function(x) {
+  if (length(x$params) == 0L)
+    return(x)
```
jennybc: Ditto about the extra careful use of brackets

R/workhorse.R
LucyMcGowan: oh yikes, I can switch to parsing myself, I hadn't noticed that disclaimer 🙈

R/workhorse.R
LucyMcGowan: 😱 I used the "Reformat Code" in RStudio and it changed those (& I didn't notice) do you generally just use the "Reindent Lines"?

R/workhorse.R
LucyMcGowan: okay I've switched it to use jsonlite 🎉

jennybc:
If I just focus on the PR's title (add check for content), I think the is almost done. But can you add some positive and negative tests to make sure the new check is doing it's job?

Something that still catches my eyeballs but is better resolve elsewhere is this:

```r
if (grepl("^http", x$endpoint)){
    x$url <- x$endpoint
  } else{
    x$url <- file.path(x$api_url, sub(".", "", x$endpoint))
  }
```

It feels like we should be using `httr::modify_url()` here and, to the extent that we cannot, it implies the ingredients are being stored in the most natural form. Also not sure what the regex substitution is about. But, again, I think this is a conversation we're having elsewhere about generalized vocabulary.

LucyMcGowan:
re: using `httr::modify_url()` all of those changes are in PR #20, so I haven't touched them elsewhere, but yes I agree

jennybc:
OK so then you just need to see if you're doing all you can test-wise. This is your best chance to get out of ahead of such things. Otherwise you just discover bugs / test cases the hard way and have to come back, once you've mentally moved on.

LucyMcGowan:
I've added some tests to make sure it's doing what we've hoped. What do you think of [this logic](https://github.com/tidyverse/googledrive/blob/check_content_type/R/workhorse.R#L94-L97)? It's a little clunky, but it lets me test the function (or parts of it) without calling the API (I have API tests too, but I figure it's good to have a few that can test on travis etc)

R/workhorse.R
```diff
@@ -75,22 +75,24 @@ make_request <- function(x, ...){
...
-          "\n",
-          "Actual content-type:\n%s"
-        ),
-        expected,
-        actual
-      )
-    }
-    jsonlite::fromJSON(httr::content(res, "text"), simplifyVector = FALSE)
+                            expected = "application/json; charset=UTF-8",
+                            internet = TRUE) {
```
jennybc: What's the rationale for an `internet` argument here?

If it's testing related, which I'm guessing it is, let's see if we can find an approach that serves the same purpose that is contained to the tests.

LucyMcGowan: 👍  agree that is what I was referring to in the comment above:
> What do you think of this logic? It's a little clunky, but it lets me test the function (or parts of it) without calling the API (I have API tests too, but I figure it's good to have a few that can test on travis etc)

tests/testthat/test-process_request.R
```diff
@@ -0,0 +1,86 @@
+test_that("process_request properly errors on wrong content type", {
```
jennybc: You might want to split these into a smaller number of tests. Why? so that your search is narrower when/if things fail.

tests/testthat/test-process_request.R
```diff
@@ -0,0 +1,86 @@
...
+  skip_on_appveyor()
+  skip_on_travis()
+
+  ## something we know won't be correct
+  res <- httr::GET("http://livefreeordichotomize.com")
+
+  expected = "application/json; charset=UTF-8"
+  actual = res$headers$`content-type`
+  ## let's make sure our test is actually good, this should be true
+  expect_true(actual != expected)
```
jennybc: I would probably use `expect_false()` on the outside and an explicit equality-testing function inside, such as `identical()` or `all.equal()` with arguments appropriate to your expectation.

tests/testthat/test-process_request.R
```diff
@@ -0,0 +1,86 @@
...
+  expect_error(process_request(res),
+               sprintf(paste0("Expected content-type:\n%s",
+                              "\n",
+                              "Actual content-type:\n%s"
+               ),
+               expected,
+               actual)
+  )
+
+  x <- drive_list()[1,] # grab most recent thing on drive
```
jennybc: From here to the end, it feels like we're doing unrelated stuff. Is there a close connection?

LucyMcGowan: I was trying to prove that it works with a generic call from our api, but maybe that is silly 😬, I can take out that last part

R/workhorse.R
jennybc: My main confusion is this: the function we're talking about, `process_request()` never calls the internet. So why would it have an internet argument?

LucyMcGowan: hmm yes that makes sense. In order to get an object that `httr::stop_for_status()` and `httr::content()` won't error on, I was thinking we needed the internet (unless we save some object that httr recognizes as a response?)

jennybc: If the problem is test input, i.e. it's not a full-fledged httr response object, here's an option:

Come up with a call that creates such an object with the properties you need. Use it to capture a response. Comment it out. Save this as a "reference object" and use `testthat::expect_equal_to_reference()`. Or just generally load as needed, in order to test response processors.

tests/testthat/test-process_request.R
jennybc: That might be a good thing to do, but in a separate test. And with a reference object you've stored (see above).

R/workhorse.R
jennybc: We posted at the same time but see above for advice about making and storing an object to test with.

Issue #19, opened 2017-05-16
Mimetype table


LucyMcGowan:
* Store this in .drive environment.
* Centralize all info needed to act this way versus that based on mime type.

LucyMcGowan:
@jennybc I've [started this](https://github.com/tidyverse/googledrive/blob/mimeType_tbl/data-raw/mime-types.R) & created the tables, how do I stick them in the `.drive` environment within the package? Do I save as .rdata & load into?

jennybc:
Minor point: If I follow correctly, `zoo::na.locf(.)` could probably be replaced with `tidyr::fill()`, to stay more within tidyverse.

Here's an example of me [using `devtools::use_data(..., internal = TRUE)`](https://github.com/tidyverse/googlesheets2/blob/90ba56c18c11631a28b532b66c0c708776e8f549/data-raw/discovery-doc-prep.R#L188) to make a list named  `.endpoints` available to package functions. That is one option. But it does not put the object into any particular environment.  The `.drive` environment does seem like a logical place to store this.

Also, if you want to stash multiple objects, I think you would need to have them all available and do that at once. So this is possibly not the best approach in this case?

I suspect you should have an explicit assignment of this information to a tibble in `.drive`. It's great that you already have a data fetching/preparation script in `data-raw`. I suggest you write the results to csv below `inst/extdata` and load this to an internal data frame. About to push a commit to the relevant branch with a demo. You'll need to adapt to you scenario.

LucyMcGowan:
* use the mime type table to intelligently choose upload file type

jennybc:
Did you mean to close this when you merged #28?

LucyMcGowan:
😱 yes!

Issue #7, opened 2017-05-11
Interface of drive_list()


jennybc:
I think you should make heavy reference to `list.files()`. Specifically:

  * `path`: Should default to meaning "list my Drive files", which strikes me as the closest match for "current working directory". But then user should also be able to specify a *folder* here as well.
  * `pattern`: that's what the regex should be called, instead of `search`
  * I think `type` should possibly be elevated to an argument in the signature of `drive_ls()` itself, as opposed to getting mopped up in `...`.
  * Lots of the other args of `list.files()` are also worth considering. Good to ask yourself: is there is an obvious translation of this into the world of Drive?

In general, you need to think about how to help people craft searches. Obviously step one is simply to link to the [file searching guide](https://developers.google.com/drive/v3/web/search-parameters), as we've done. And to include some good examples. But this is an example of something that might be ripe for an add-in down the road.

jennybc:
Now that I actually understand this a bit better, these seem like areas to explore:

  * Forming `q`. At its simplest, people have to provide `q` verbatim. But it might be nice to accept a vector of `q`s or multiple `q`s in `...`, which `drive_list()` would then concatenate with `and`.  It feels like there's also a need for an internal "append this to `q`" helper.
  * Specifying the fields that are extracted into columns of the returned tibble. It feels like this should have a reasonable set of defaults (the current hard-wired ones?), but be exposed for the user to tweak.
  * Elevate `type` to an actual argument? repeated from above.

LucyMcGowan:
from #21 

context: user is interested in the file `baz` in the folder `foobar` and would have to type
```
drive_list(path = "foobar", pattern = "baz")
```
>as a user, why can't I just provide foobar/baz? What would be the difference? 🤔 I suspect answering this has implications for drive_list() docs and/or logic. Maybe put this in the existing issue thread for drive_list() interface? It has nothing to do with the README.

@jennybc 🤔 that is a good point, the way we have the logic now we expect the path parameter to always be a folder. In the broad sense, the pattern would usually be something like you'd want to find all the files with baz in them in a specific folder? But I can see a user wanting to specify a specific file and being sad that this doesn't let you do that with just one parameter. But then also if you are trying to find a specific file is that the best use of a "list"? (as you can see I've talked myself right around in a circle here). If keeping with list.files I don't believe it lets you pass a full path to the first parameter, it expects a directory and lists everything inside it.

jennybc:
[...]

PR #28, opened 2017-05-19
Create mime type tibble for .drive (#19)

LucyMcGowan:


R/aa_utils.R
```diff
@@ -12,11 +12,22 @@ NULL
 .drive <- new.env(parent = emptyenv())
 .drive$base_url_files_v3 <- "https://www.googleapis.com/drive/v3/files"
 .drive$base_url <- "https://www.googleapis.com"
+.drive$mimetypes <-
+  system.file("extdata", "mimetypes.csv", package = "googledrive") %>%
+  read.csv(stringsAsFactors = FALSE)
```
jennybc: You don't mean to keep my toy example in here, do you? If not, these lines and the file they load should go away.

LucyMcGowan: yikes no I didn't mean to keep that - I accidentally merged it in with my last commit 🙈

LucyMcGowan:
okay I think this is ready for another look @jennybc. I 
* updated the mime type tibble to join in the extensions from the `mime::mimemap` & incorporated that in the `drive_upload()` if the type was not specified. Right now, in terms of specifying type, I only let it be a document, spreadsheet, or presentation (as it was before, I didn't change it [per this comment](https://github.com/tidyverse/googledrive/issues/7#issuecomment-302839031)) 
* updated some testing according
* Noticed that you cannot change the publication status of "non-Google Drive" files (files without "application/vnd.google-apps." in their mimeType), which makes sense because google doesn't track revisions with these files and publishing is an action on a revision ID, I added an error accordingly if the file doesn't have a Google mimeType

jennybc:
This basically looks good to me. Do you want to deal with the conflicts then merge? I think you should remove the fiddling with, say, `drive_list()` tests from this PR, so this is really just about mime types. Then it should merge cleanly. The other will need to be revised in light of #34 anyway and should be a separate PR or commits.

Re: our awful testing situation. Maybe each file that hits the API should have the code necessary to create the structures at the top, inside `if (FALSE) { ... }`. Then it's clearly filed with the relevant tests and either of us can rerun it as necessary.

Issue #24, opened 2017-05-16
Path handling toolkit


jennybc:
I have a second generation solution for resolving a requested path, such as `foo/bar/baz`.

I'd like to look ahead a bit and package it more generally, i.e. for use beyond `drive_ls()`.

@LucyMcGowan Will you help me list the ways other functions will need to work with paths?

`drive_list()`: User can supply path. We need to make sure such a path exists and is uniquely defined. Then determine file id for the terminal leaf and list it.

Functions that create or move files: We need to determine the maximal existing stem, in case some folders need to be created. Example: user asks to write a file to `foo/bar/baz/yo/myfile.txt`. We determine that `foo/bar` exists and that we need to create `baz` within `foo/bar` and `yo` within `foo/bar/baz`.

What else do we need?

LucyMcGowan:
Those are the only two we currently use it for - off the top of my head I can't think of other use cases for needed to find a folder other than determining whether a path exists, if not, can we make it, it so, is it unique.

jennybc:
OK great. Let us think about interface. A path goes in. What comes out? What do we need to do with the result?

It would be nice to have several small functions, but I think the API call burden would be too high. So I think we need one primary function that returns a list. What do we need to put into this list?

  * `n_paths` = How many such paths exist?
    - 0: path does not exist
    - 1: path exists and is uniquely defined
    - 2: more than one path exists, i.e. there are 2 or more `foo/bar/baz` paths that differ by at least one file id
  * `leaf_unique` is `TRUE` if path exists (possibly more than 1 such path!) and uniquely identifies the terminal leaf, `FALSE` otherwise
  * a list with one component per path or a tibble with one row per path?
  * a ?tibble? with one row per element of the path (split up) and info on (cumulative?) existence

Things get increasingly foggy as I go on. We need to figure what form of reporting will put as much logic as possible in this helper and list, so that our calls elsewhere are short and readable. Once I can take a first pass, we'll try it out and probably need to iterate a couple time.

LucyMcGowan:
Would the list (or tibble) with one component per path contain the leafmost id?

jennybc:
The leafmost id appears in the last two of the proposed objects up there. But not in a terribly accessible form. So yes that should be positioned for easier access.

In general, there's no guarantee there is exactly one leafmost id, right? I don't think this helper should throw many errors. I think the callers (or wrappers) have to use the return value to figure out whether to error.

[...]

PR #34, opened 2017-05-21
Recursive drive path helper; relates to #7 and #24

jennybc:
#7 Interface of drive_list()
#24 Path handling toolkit

This is still WIP, so not quite done.

@LucyMcGowan Do you want to have a look around and discuss? I'm especially interested to know if you see anything where you feel this is a step *backwards*.

I still need to add more tests but the whole testing setup is very frustrating right now, so am also holding off a bit. I think we need to create some directory structures that are "stable", so our tests are faster. If we aren't testing a function that edits Drive, the test should not edit Drive. We should have the  content ready and waiting to test against.


LucyMcGowan:
I agree re:testing - I was originally trying to set it up so someone who didn't have access to our testing account could run everything on their own account, but it seems very superfluous to recreate these folders everytime. Maybe I should write up a script that creates them all *one time* & have it commented out in the test folder like we have other file generating scripts?

LucyMcGowan:
Also, I ended up changing the tests in the mime type PR so that it created unique folders so they won't fail for those reasons, so I'll set up the general structure that way as well.

LucyMcGowan:
just took it for a quick spin & it looks great! I'm going to revamp tests a bit today to be less of a huge pain (but maybe we can make a robust plan on Wednesday)

jennybc:
I'm trying to rescue the appveyor build. I think I'll merge, so it's doesn't get too large and we don't have too many things evolving separately. About to look at your mime table PR again.

PR #41, opened 2017-05-21
fix #40 drive_mkdir allow for multiple folders with the same name 

### LINKED ISSUE #40: drive_mkdir doesn't allow for multiple folders with the same name in the same path
as mentioned in #24 
I think there may be a small bug in the get_leaf function - when I run:
```
drive_mkdir("baz", path = "test_8675309/foo/yo")
```
I see
```
## Error: The path 'test_8675309/foo/yo/' identifies more than one file:
## File of type folder, with id 0B0Gh-SuuA2nTOEVQX1BteW1Fa3M.
## File of type folder, with id 0B0Gh-SuuA2nTN3VMYWdOZlhrM0k.
```
but on drive I have:
test_8675309
+-- foo
| +-- yo (id 0B0Gh-SuuA2nTOEVQX1BteW1Fa3M)
+-- yo (id 0B0Gh-SuuA2nTN3VMYWdOZlhrM0k)

so in fact test_8675309/foo/yo should be unique (since the other one that it's fussing about is actually test_8675309/yo)

Also, when checking through the path, match [here](https://github.com/tidyverse/googledrive/blob/master/R/drive_path.R#L33) doesn't allow for nested folders with the same name. For example if we had "foo/foo", which is perfectly logical, match only matches the first, so we would have depth 1 for both, where they should each be listed as both depth 1 and 2.

LucyMcGowan:


R/drive_path.R
```diff
@@ -30,7 +30,13 @@ get_leaf <- function(path = NULL) {
   ## guarantee: we have found at least one item with correct name for
   ## each piece of path
 
-  hits$depth <- match(hits$name, path_pieces)
```
jennybc: Yeah, this is *alleged* depth, not actual. But then below I need to use the information when vetting the paths-to-root. Currently I'm not doing so, which leads to the false positive you discovered. ☹️ I will probably make a separate PR with an alternative fix.

LucyMcGowan: sounds good! 👍

PR #43, opened 2017-05-22
Bug fix get_leaf(); fixes #40

### LINKED ISSUE #40: drive_mkdir doesn't allow for multiple folders with the same name in the same path

jennybc:
This is a fix for #40 based on path length.

I still need to address this point from #40 and will borrow your approach from #41.

> Also, when checking through the path, match here doesn't allow for nested folders with the same name. For example if we had "foo/foo", which is perfectly logical, match only matches the first, so we would have depth 1 for both, where they should each be listed as both depth 1 and 2.

### DIFF
--- a/R/drive_path.R
+++ b/R/drive_path.R
@@ -127,10 +127,6 @@ form_query <- function(path_pieces, leaf_is_folder = FALSE) {
   glue::collapse(c(leaf_q, dirs_q), last = " or ")
 }
 
-require_length <- function(paths, len) {
-  paths %>% purrr::keep(~ length(.x) == len)
-}
-
 rootwise_parent <- function(paths) {
   ## retain only paths that end with root_id <==> not NA
   paths <- paths %>% purrr::keep(~ !is.na(last(.x)))
--- a/R/drive_path.R
+++ b/R/drive_path.R
@@ -39,6 +39,9 @@ get_leaf <- function(path = NULL) {
   ## for each candidate, enumerate all upward paths, hopefully to root
   root_path <- leaf_id %>%
     purrr::map(pth, kids = hits$id, elders = hits$parents, stop_value = root_id)
+  ## retain candidate paths with d + 1 elements = d path pieces + root_id
+  root_path <- root_path %>%
+    purrr::map(require_length, len = d + 1)
   ## for each candidate, get an immediate parent on a path back to root
   ## will be NA is there is no such path
   root_parent <- root_path %>%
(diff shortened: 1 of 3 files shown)


jennybc:
> I think this will still cause a problem

See what you think now. I reworked things a bit.

R/drive_path.R
```diff
@@ -30,39 +30,44 @@ get_leaf <- function(path = NULL) {
   ## guarantee: we have found at least one item with correct name for
   ## each piece of path
 
-  hits$depth <- match(hits$name, path_pieces)
-  hits <- hits[order(hits$depth), ]
-
   ## leaf candidate(s)
-  leaf_id <- hits$id[hits$depth == d]
+  leaf_id <- hits$id[hits$name == last(path_pieces)]
```
jennybc: I think this is going to generalize better to case where we need to loop backwards over path pieces, until we find one that actually exists.

LucyMcGowan: oh lovely! it works! 🎉

jennybc:
Please have one last look.I removed a lot of tests that seemed redundant. I think this fixes the problems you spotted (foo/bar vs bar, foo/foo) and at least one new one I discovered along the way (foo/bar/baz vs bar/foo/baz). If you think looks ok, let's merge.

LucyMcGowan:
looks good to me 👍

PR #47, opened 2017-05-22
update readme & demos with new functions

LucyMcGowan:


R/drive_publish.R
```diff
@@ -19,41 +19,42 @@ drive_publish <- function(file = NULL,
...
+  )
   response <- make_request(request, encode = "json")
   proc_res <- process_drive_publish(response = response,
                                     file = file_update,
                                     verbose = verbose)
 
-  ##for some reason we have to make the request 2x?
+  ## if we want to autopublish, it must have already been published, so
+  ## we need to run again
+  if (x$published == TRUE & x$publishAuto == TRUE) {
```
jennybc: If `x$published` and `x$publishAuto` are logical, which I assume they are, this should just be `x$published && x$publishAuto`. Note the use of `&&`. If you are worried about NAs, then you could do `isTRUE(x$published) && isTRUE(x$publishAuto)`.

R/drive_publish.R
```diff
@@ -94,23 +95,15 @@ drive_check_publish <- function (file = NULL, verbose = TRUE) {
...
-  response <- make_request(request)
-  proc_res <- process_request(response)
-
-  last_rev <- length(proc_res$revisions)
-  rev_id <- proc_res$revisions[[last_rev]]$id
-
   fields <- paste(c("id", "published", "publishAuto", "lastModifyingUser"),
                   collapse = ",")
 
+  revisionId = "head"
```
jennybc: `<-` ?

README.Rmd
```diff
@@ -99,6 +97,26 @@ drive_chickwts
 ```
 Much better!
 
+#### Publish files
+
+Revisions of Google Documents, Sheets, and Presentations can be published online. By default, `drive_publish()` will publish your most recent revision. Running this function will update a tibble in
+your `gfile` object with publication information. You can also check your publication status by running `drive_check_publish()`.
+
```
jennybc: Do you think more people will connect with the word "revision" or "version"? I know "revision" is more technically correct, but suspect "version" resonates with more people. I would remove the sentence about updating a tibble in the `gfile` object, since the typical README reader won't want to know the internal stuff. I like the ability to check the publishing status! 💪

internal/02_sharing_demo.md
```diff
@@ -24,27 +25,26 @@ drive_upload("demo.txt", "Happy Little Demo")
 
     ## File uploaded to Google Drive: 
     ## demo.txt 
-    ## As the Google document named:
+    ## As the Google text/plain named:
```
jennybc: Sidebar: once we have human-friendly nicknames stored for the mime types, this will be a great place to use them.

internal/03_publish_demo.md
```diff
@@ -3,86 +3,136 @@ publish
...
+gs_auth("sheets-token.rds")
 ```
 
-    ## The latest revision of the Google Drive file 'Happy Little Demo' is not published.
+this shouldn't work
+-------------------
 
 ``` r
-my_file$publish
+try(gs_url(url, lookup = FALSE))
```
jennybc: I think you should possibly be setting visibility to private here. I mean, we will still want to demonstrate an error, but for the right reasons.

internal/03_publish_demo.md
```diff
@@ -3,86 +3,136 @@ publish
...
-    ##                <dttm>    <chr>     <lgl>        <lgl>           <chr>
-    ## 1 2017-05-11 10:47:13        3     FALSE        FALSE Lucy D'Agostino
+    ## Sheet-identifying info appears to be a browser URL.
+    ## googlesheets will attempt to extract sheet key from the URL.
 
-publish it
-----------
+    ## Putative key: 1S-GYudy2qu_JMZCVPYwKUH2Kh3hLUeqtwH7-XrF_Vs8
+
+    ## Worksheets feed constructed with public visibility
```
jennybc: This is why I know the visibility is still set to public.

internal/03_publish_demo.Rmd
```diff
@@ -6,54 +6,73 @@ output:
...
+drive_check_publish(drive_chickwts)
 ```
 
-now we have tibble with publication status
-
+## try again!
 ```{r}
-my_file$publish
+gs_url(url, lookup  = FALSE)
 ```
```
jennybc: Do you have a definitive answer for this: is it possible for a Sheet to NOT be shared with the world but accessible via the API? I would dearly love to have a matrix of sharing statii vs publication statii and, for each cell, to understand if it's possible. This isn't really your problem, but you'd be my hero if we could get to the bottom of it and document that.

LucyMcGowan: Yes (I think) - so this file is an example of one that is accessible through the API, since it's published, but not shared with the world via link (in a browser). I can add the sharing permissions to the doc to show that as well. If I try to access this link in a browser (with my other google account) it claims I don't have permissions to.

LucyMcGowan:
I think if I understand what you're asking here: 
>Do you have a definitive answer for this: is it possible for a Sheet to NOT be shared with the world but accessible via the API? 
the answer is yes! the example I have in this demo shows a file that is shareable through the API (after I publish it, I can see it with my other google account) but not through a link in a browser. This seems a bit counter intuitive since all I am using to access it via the API _is_ the link, but I think it is the case. In the updated I just pushed, I included the [url](https://docs.google.com/spreadsheets/d/1dES65Ur8AcYSS4JnnMGRp7A3wVJEHKMvBG8M60U4m0A/edit?usp=drivesdk) & I commented out cleaning up so the file should still exist on our tidyverse.testdrive account so you can test it yourself with your own google account to double check that it is not visible via the browser.

internal/03_publish_demo.Rmd
jennybc: OK that seems pretty definitive. I can't walk through it myself right now but look forward to doing so.

jennybc:
> since all I am using to access it via the API is the link

This isn't as weird as you think, because the URL is really only being used to extract file id.

jennybc:
I'm so excited about the publishing, both the ability to check and modify! 🎉

PR #50, opened 2017-05-23
remove map_df (fixes #35)

### LINKED ISSUE #35: Avoid use of purrr::map_df()
It turns out `purrr::map_df()` does still call dplyr, so if we want to remove our dplyr dependency, there are various functions in purrr we should avoid. This is one of them.

https://github.com/tidyverse/purrr/issues/328

LucyMcGowan:
this is a bit clunky, but it gets rid of our dplyr dependency 🎉

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,7 +1,7 @@
 ^.*\.Rproj$
 ^\.Rproj\.user$
 ^internal$
-^\.httr-oauth$
+^.*\.httr-oauth$
 ^\.travis\.yml$
 ^README\.Rmd$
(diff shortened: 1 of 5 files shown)


jennybc:
I don't have any example input to play with. But it seems like you could still use `tibble::add_row()` here, given that you create a tibble with all the necessary variable first, regardless. Otherwise, I think `modifyList()` might be a way to eliminate the need for `fill_tbl()`. Finally, maybe you could build the tibble up variable-wise and use `.null` to deal with missing pieces? If any one this unclear, let's look at it in person.

PR #51, opened 2017-05-23
update docs to reflect drive_upload change (#28)

LucyMcGowan:
when I added the mime type tibble, I changed how `drive_upload()` determines the file type (it used to default to a Google Drive type, and now defaults to the type in the tibble) - I forgot to update the documentation to reflect this change.

### DIFF
--- a/R/drive_upload.R
+++ b/R/drive_upload.R
@@ -5,14 +5,9 @@
 #'   the file
 #' @param overwrite logical, do you want to overwrite file already on Google
 #'   Drive
-#' @param type if type = `NULL`, will force type as follows:
-#' * **document**: .doc, .docx, .txt, .rtf., .html, .odt, .pdf, .jpeg, .png, .gif,.bmp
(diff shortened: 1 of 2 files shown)


R/drive_upload.R
```diff
@@ -5,14 +5,9 @@
...
-#' * **document**: .doc, .docx, .txt, .rtf., .html, .odt, .pdf, .jpeg, .png, .gif,.bmp
-#' * **spreadsheet**: .xls, .xlsx, .csv, .tsv, .tab, .xlsm, .xlt, .xltx, .xltm,
-#'   .ods
-#' * **presentation**: .opt, .ppt, .pptx, .pptm
-#'
-#'  otherwise you can specify `document`, `spreadsheet`, or `presentation`. Files with no extension will
-#'   be assumed to be a `folder`
+#' @param type if type = `NULL`, will upload as a non-Google Drive document
+#'  otherwise you can specify `document`, `spreadsheet`, or `presentation`.
+#'  Files with no extension will be assumed to be a `folder`.
```
jennybc: I wonder if this is a good default. Seems like it might be a bit over-eager on folder creation. Or do we basically have no choice, i.e. there is no better fallback type?

PR #54, opened 2017-05-25
fix tests to make all files in advance

LucyMcGowan:
@jennybc here is a first pass at this :tada:

### DIFF
--- a/tests/testthat/test-drive_list.R
+++ b/tests/testthat/test-drive_list.R
@@ -8,26 +8,94 @@ context("List files")
 # (pesky_files <- drive_list(pattern = "foo|bar|baz|yo"))
 # pesky_files$id %>% purrr::map(drive_file) %>% purrr::map(drive_delete)
 
+nm_ <- function(x) paste0(x, "-TEST-drive-list")
+
+run <- FALSE
+clean <- FALSE
+if (run) {
+  ## make sure directory is clean
+  if (clean) {
+    del_ids <- drive_list(pattern = paste(c(nm_("foo"),
+                                            nm_("foo2"),
+                                            nm_("bar"),
+                                            nm_("baz"),
+                                            nm_("yo")),
+                                          collapse = "|")
+    )$id
+    del_files <- purrr::map(del_ids, drive_file)
+    del <- purrr::map(del_files, drive_delete, verbose = FALSE)
+  }
+
+  ## drive_list() not confused by same-named folders
+  ## get_leaf() errors when two distinct folders have same path
+  ## +-- foo
+  ## | +-- bar
+  ## | +-- bar
+  ## | | +-- baz
+  ## | | | +-- yo
+
+  drive_mkdir(nm_("foo"), verbose = FALSE)
+  drive_mkdir(nm_("bar"), path = nm_("foo"), verbose = FALSE)
+  drive_mkdir(nm_("baz"), path = file.path(nm_("foo"), nm_("bar")), verbose = FALSE)
+  drive_mkdir(nm_("yo"), path = file.path(nm_("foo"), nm_("bar"), nm_("baz")), verbose = FALSE)
+  drive_mkdir(nm_("bar"), path = nm_("foo"), verbose = FALSE)
+
+  ## get_leaf() is not confused by same-named leafs at different depths
+  # +-- foo
+  # | +-- bar
+  # | | +-- baz
+  # +-- baz
+  drive_mkdir(nm_("baz"), verbose = FALSE)
+
+  drive_mkdir(nm_("foo2"), verbose = FALSE)
+
+  ## get_leaf() is not confused by nested same-named things
+
+  # +-- foo (folder)
+  # | +-- foo (document)
+
+  write.table(chickwts, "chickwts.txt")
+  drive_upload(input = "chickwts.txt",
+               output = file.path(nm_("foo"), nm_("foo")),
+               verbose = FALSE)
+
+  ## get_leaf() is not confused by differently ordered non-leaf folders
+  # +-- bar
+  # | +-- foo
+  #   | +-- baz (file)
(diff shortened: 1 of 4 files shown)


tests/testthat/test-drive_list.R
```diff
@@ -8,97 +8,133 @@ context("List files")
 # (pesky_files <- drive_list(pattern = "foo|bar|baz|yo"))
 # pesky_files$id %>% purrr::map(drive_file) %>% purrr::map(drive_delete)
 
+nm_ <- function(x) paste0(x, "-TEST-drive-list")
```
jennybc: This will soon move over to helper.R, for general use.

tests/testthat/test-drive_list.R
```diff
@@ -8,97 +8,133 @@ context("List files")
...
+run <- FALSE
+clean <- FALSE
+if (run) {
+  ## make sure directory is clean
+  if (clean) {
+    del_ids <- drive_list(pattern = paste(c(nm_("foo"),
+                                            nm_("foo2"),
+                                            nm_("bar"),
+                                            nm_("baz"),
+                                            nm_("yo")),
```
jennybc: `nm_()` is vectorized, so I think this can be `paste(nm_(c("foo", "foo2", "bar", "baz", "yo")), collapse = "|")`.

tests/testthat/test-drive_list.R
```diff
@@ -8,97 +8,133 @@ context("List files")
...
+  drive_mkdir(nm_("bar"), path = nm_("foo"), verbose = FALSE)
+
+  ## get_leaf() is not confused by same-named leafs at different depths
+  # +-- foo
+  # | +-- bar
+  # | | +-- baz
+  # +-- baz
+  drive_mkdir(nm_("baz"), verbose = FALSE)
+
+  drive_mkdir(nm_("foo2"), verbose = FALSE)
```
jennybc: Do we have a reason for making this folder?

tests/testthat/test-drive_list.R
jennybc: There seem to be more instances below where this may also apply.

tests/testthat/test-drive_list.R
jennybc: Oh, I see you're making the slug specific to this file, but also short. You could define a function for making a function that adds a slug in helper.R and then call it in each file. It doesn't really matter. We might start to care if the slugs get more complicated, i.e. indicate who's running the tests.

``` r
## in helper.R
nm_fun <- function(slug) {
  function(x) paste0(x, slug)
}

## at the top of this file
nm_ <- nm_fun("-TEST-drive-list")

## usage
nm_("hi")
#> [1] "hi-TEST-drive-list"
```

tests/testthat/test-drive_list.R
```diff
@@ -8,97 +8,133 @@ context("List files")
 # (pesky_files <- drive_list(pattern = "foo|bar|baz|yo"))
 # pesky_files$id %>% purrr::map(drive_file) %>% purrr::map(drive_delete)
```
jennybc: These leading comments will need to be deleted or modified before we merge.

tests/testthat/test-drive_list.R
```diff
@@ -8,97 +8,133 @@ context("List files")
...
-  ## clean up
-  foo_id %>%
-    drive_file() %>%
-    drive_delete()
+  expect_identical(nm_("yo"),
+                   drive_list(path = file.path(nm_("foo"),
+                                               nm_("bar"),
+                                               nm_("baz"))
+                   )
+                   $name)
```
jennybc: Or (this purely about style):

```
expect_identical(
    nm_("yo"),
    drive_list(path = file.path(nm_("foo"), nm_("bar"), nm_("baz")))$name
  )
```

tests/testthat/test-drive_list.R
```diff
@@ -8,97 +8,133 @@ context("List files")
...
+    del <- purrr::map(del_files, drive_delete, verbose = FALSE)
+  }
+
+  ## drive_list() not confused by same-named folders
+  ## get_leaf() errors when two distinct folders have same path
+  ## +-- foo
+  ## | +-- bar
+  ## | +-- bar
+  ## | | +-- baz
+  ## | | | +-- yo
```
jennybc: Do you think it would be useful to repeat these directory trees in the associated tests?

tests/testthat/test-drive_list.R
```diff
@@ -8,97 +8,133 @@ context("List files")
...
-  bar_2_id <- drive_mkdir("bar")$id
-  expect_identical(get_leaf("foo/bar/")$id, bar_id)
-
-  c(foo_id, bar_2_id) %>%
-    purrr::map(drive_file) %>%
-    purrr::map(drive_delete)
+  expect_length(get_leaf(file.path(nm_("foo"),
+                                   nm_("bar"),
+                                   nm_("baz"))),
+                3)
```
jennybc: The original test didn't involve `baz`. Something was lost in translation here.

`expect_silent()` is also an option.

LucyMcGowan: I think I changed it to `baz` because I already used `bar` in that place for a different test so since I'm not deleting & recreating them each time, I had to pick a different name - but I'll double check. 👍🏼 For `expect_silence`

jennybc:
Once you read/respond to the comments, go ahead and merge. Or do you want to do for all affected files at once?

It is noticeably faster, which is nice!

LucyMcGowan:
sounds good! I think I'll go through and update them all tomorrow before I merge - very excited about the speed! 🏃‍♀️💨

jennybc:
When you revisit this one and do others, can you take a really hard look for simplicity and redundancy? This is a good chance to eliminate tests that are doing the same thing.

PR #58, opened 2017-05-28
Non-judgmental, testable path finding; closes #24

### LINKED ISSUE #24: Path handling toolkit

jennybc:
@LucyMcGowan Will you take a look? I think this is ready to merge. No doubt there may be some refinements once we start using the partial paths, but I think this is a big improvement re: the path situation.


R/drive_path.R
```diff
@@ -1,91 +1,193 @@
...
 
-  if (length(keep_me) > 1) {
-    line0 <- glue::glue("The path '{path}' identifies more than one file:")
-    lines <- glue::glue_data(
-      hits[hits$id %in% leaf_id[keep_me], ],
-      "File of type {type}, with id {id}."
-    )
+  ## ensure path_orig gets added, even if there are zero rows
+  leaf_tbl$path_orig <- rep_len(path, nrow(leaf_tbl))
+  leaf_tbl
```
jennybc: Once we revisit the object that holds file metadata (maybe becoming a dribble?), that should probably be what this returns. `drive_path()` is an exported function that exposes this function and it should probably drop the `path_orig` variable. I figure we can refine this when we create the dribble class and methods.

LucyMcGowan:
just took it for a spin -- looks awesome! 🎉

PR #59, opened 2017-05-28
Make possible to list top-level files in My Drive; fixes #46

### LINKED ISSUE #46: Make it easy to get top-level listing of My Drive
I now realize that `drive.files.list` lists My Drive, by default, *without regard to folder hierarchy*. I put this in `drive_list()`:

``` r
  ## if path reduces to root (i.e., "My Drive"), make it an explicit NULL
  path <- rationalize_path(path)
```

which should be revisited. I now think `path = "~"` or `path = "~/"` or `path = "/"` should get a listing of files with root as direct parent. Which is different from `path = NULL`.

jennybc:


jennybc:
Closing, now part of #58

PR #70, opened 2017-05-31
WIP: re-fashion drive_list() as drive_search() (closes #69 and closes #53)

### LINKED ISSUE #53: drive_upload overwrite should check if the name is unique
I think this can be solved by changing
``` 
   old_doc <- drive_list(path = path,
                          pattern = paste0("^", name, "$"),
                          verbose = FALSE)
    if (!is.null(old_doc)) {
      id <- old_doc$id[1]
    }
  }
```
to 
```
    old_doc <- drive_list(path = output
                          verbose = FALSE)
    if (!is.null(old_doc)) {
      id <- old_doc$id

    }
  }
```

### LINKED ISSUE #69: drive_upload() uses the updated drive_path()/drive_list()
* currently we are using the old logic to traverse the files, but this should be replaced with the new `drive_path()`

jennybc:
Existing `drive_list()` was a messy hybrid of path finding, folder listing, searching, and filtering.

Proposal: kill existing `drive_list()`, in favour of a simpler function `drive_search()`. Its job is to do the operations described in the [Search for Files guide](https://developers.google.com/drive/v3/web/search-parameters). All path work should go through `drive_path()`. I've added an example to `drive_search()` for how to list the contents of a folder. We could wrap that nicely in future and call it `drive_list()`.

It is unavoidable to patch up `drive_upload()` because it calls `drive_list()` and it's used in many tests. I've made it dramatically simpler but have left some to do's for you. Can you finish this off so we pass tests again? The goal is to do as little as possible until we straighten out the dribble thing.


LucyMcGowan:
ok - I think it's working - the only piece that maybe I can add here is stripping the extensions if `type` isn't `NULL`?

jennybc:
I'm a bit foggy on the significance of the file extension, the implications of specifying mimeType (or not), and the interaction between the two. And what it takes to request conversion.

These seem like natural scenarios.

  * mimeType is `NULL`. File should be uploaded as a blob, i.e. binary type, and with its name un-edited. Is this currently what happens if no mimeType is given for a new file? Or does Drive attempt to detect and convert?
  * If the file is being converted to a native format (Docs, Sheets, Presentation), then I can see why one might want to strip the extension.

LucyMcGowan:
If the mimeType is `NULL` as we have it now `httr::file_upload()` determines the type. I agree that the extension should stay unless the file is being converted.

jennybc:
I think you should also re-name `test-drive_list.R` to `test_drive_search.R` and 😮 gut it. We can retrieve all that file/folder setup later, if necessary. Maybe keep just enough to write a couple basic tests of `drive_search()`. And delete my comments about all that.

jennybc:
> If the mimeType is NULL as we have it now httr::file_upload() determines the type. I agree that the extension should stay unless the file is being converted.

Can we keep httr from doing that?

LucyMcGowan:
I'm not sure how to 😕 from the docs:
> `type` mime type of path. If not supplied, will be guess by guess_type when needed.
& a little typo 🙈 guess -> guessed

jennybc:
Perhaps you should just call `curl::form_file(path, type)` directly. The only additional thing `httr::upload_file()` does is to guess the mime type, which it sounds like we would prefer to suppress.

LucyMcGowan:
besides the codecov sadness, do you think this is ready to merge?

PR #72, opened 2017-06-01
Hello dribble; closes #62, closes #35

### LINKED ISSUE #35: Avoid use of purrr::map_df()

### LINKED ISSUE #62: change `drive_file()` to output a dribble
* `drive_file()` should output something similar to a single row output from `drive_list()`
* perhaps develop a drive tibble (dribble) method
* switch to gfiles as only an internal thing

jennybc:
This is a sketch of what this should look like. It's just a bare start, but hopefully enough to demonstrate the overall plan.

This creates the dribble class, to the extent that S3 classes really exist! Then defines lots of `as.dribble()` methods to get a dribble from all the obvious inputs. This should be called on all user input that specifies a Drive file and on all output that returns Drive files. The `drive_hello()` function is a toy example so I can make this demo. This function can deal with input that is character (interpreted as path), character marked with `drive_id` class (interpreted as file id), and dribble.

``` r
devtools::load_all(".")
#> Loading googledrive
#> Auto-refreshing stale OAuth token.

drive_hello("Copy of chickwts")
#> Hi, my name is 'Copy of chickwts'!
#> My fileId starts with 1nfpIGA

d <- drive_hello(drive_id("1nfpIGAk97bj43DVmPivo5aFQFne7hWCqHUYSNEcyN7M"))
#> Hi, my name is 'Copy of chickwts'!
#> My fileId starts with 1nfpIGA

drive_hello(d)
#> Hi, my name is 'Copy of chickwts'!
#> My fileId starts with 1nfpIGA
```

checklist:
- [x] `drive_browse()` input + output
- [x] `drive_delete()` input
- [x] `drive_get()`  input + output
- [x] `drive_ls()`  input 
- [x] `drive_mkdir()` output
- [x] `drive_mv()` input + output
- [x] `drive_path()` input + output
- [x] `drive_publish()` input + output
- [x] `drive_search()` input + output
- [x] `drive_share()` input + output
- [x] `drive_share_link()` input + output
- [x] `drive_upload()` input + output
- [x] change `as.dribble()` ➡️ `as_dribble()`
- [x] change `file_resource` ➡️ `files_resource`
- [x] simplify test setup and tear down, in light of dribbles


R/dribble.R
```diff
@@ -0,0 +1,49 @@
+dribble <- function() {
+  structure(
+    tibble::tibble(
+      name = character(),
+      id = character(),
+      drive_file = list()
+    ),
+    class = c("dribble", "tbl_df", "tbl", "data.frame")
+  )
+}
```
jennybc: This is sort of handy when you want to return a dribble with zero rows. It is also an obvious place to put (internal) documentation about what the columns are.

R/dribble.R
```diff
@@ -0,0 +1,49 @@
...
+as.dribble <- function(x, ...) UseMethod("as.dribble")
+
+as.dribble.dribble <- function(x, ...) x
+
+as.dribble.character <- function(x, ...) {
+  structure(
+    drive_path(x),
+    class = c("dribble", "tbl_df", "tbl", "data.frame")
+  )
+}
```
jennybc: This is definitely not correct, because `drive_path()` doesn't even return the file name at the moment! Just indicative.

jennybc: All these explicit calls to `structure()` here will go away once functions like `drive_path()`, `drive_get()` etc. are brought into line and are returning dribble.

LucyMcGowan:
This looks lovely! Should I start fleshing it out tomorrow? I was looking into adding things like `].dribble` etc. so they retain the class if users do their own filtering.

jennybc:
Yes please! I think there's quite a bit to do before we get to indexing, but yes that will definitely be needed as well.

R/dribble.R
```diff
@@ -47,3 +53,27 @@ as.dribble.list <- function(x, ...) {
...
+    dribble
+  } else {
+
+    mp <- list(character = purrr::map_chr,
+               numeric = purrr::map_dbl,
+               list = purrr::map,
+               logical = purrr::map_lgl
+    )
+
+    cl <- class(dribble$drive_file[[1]][[pull]])
```
jennybc: What if the class is not one of those? I'm not sure if this comes up but we need to make sure. As discussed elsewhere, long-term, we should probably use our knowledge of the API to determine how to elevate a specific element in `drive_file` into it's own column.

LucyMcGowan: From what I've seen so far, I think the only classes that the API returns are those (and actually just 3, character, logical, and list), but I agree we should be careful.

R/dribble.R
```diff
@@ -47,3 +53,27 @@ as.dribble.list <- function(x, ...) {
     class = c("dribble", "tbl_df", "tbl", "data.frame")
   )
 }
+
+## this let's us pull things out of drive_file column that we'd like
+## as a column in the main dribble
+pull_into_dribble <- function(dribble, pull) {
+
+  if (nrow(dribble) == 1) {
```
jennybc: I don't see why a dribble with 1 row needs special handling.

LucyMcGowan: oh oops, that was from a previous iteration of the function I forgot to delete I'll remove it 🙈

R/dribble.R
jennybc: I think nevermind -- I see `list` there now. This probably does handle everything.

jennybc:
I can't review specifics until later, but here's one global thing that would be nice:

Adopt a convention for an argument that specifies one (or more?) drive file(s). We need to settle on default argument name (`drive_file`?) and an English phrase (Drive file?). Then use it everywhere possible. Write a roxygen template for this and use it religiously. This will eliminate lots of duplication and make it easy to change things later. *I think this would mean the `drive_file` column of dribble should be renamed to `files_resource`.*

Here's a relevant example from googlesheets:

https://github.com/jennybc/googlesheets/blob/master/man-roxygen/ss.R

Do similar for dribble. Use that for everything that returns dribble.

LucyMcGowan:
sounds good! do you think we should distinguish the name in the input if it can take more than 1 `dribble` (for example, `drive_mv()`, `drive_publish()`, and `drive_share()` all need a `dribble` with just 1 row, whereas `drive_delete()`, `drive_get_share_link()`, etc. can have a `dribble` with many rows)

jennybc:
For now, why don't you word the template so that it raises the possibility that some functions might require a 1-row dribble and others might allow > 1 rows. In the doc for the function itself, say that it works on one Drive file. Once the dust settles and we know which functions are vectorized and which are not, we can refine. But for input, remember, we need a template that refers to a **Drive file**, which could come as name/path, marked file id, or dribble.

R/drive_id.R
```diff
@@ -0,0 +1,10 @@
+#' Register Drive id
+#'
+#' This addes the class `drive_id` to a character string.
```
LucyMcGowan: typo 🙈

R/drive_mkdir.R
```diff
@@ -9,13 +9,27 @@
...
+      if (!is_folder(path)){
+        stop(
+          glue::glue_data(path, "'path' is not a folder:\n{name}"),
+          call. = FALSE
+        )
+      }
+    } else {
+      path <- append_slash(path)
+      path <- get_one_path(path)
+      if (path$mimeType != "application/vnd.google-apps.folder") {
```
LucyMcGowan: @jennybc do you think this makes sense? this keeps the general framework of the functions & adds in a check for whether the input is a `dribble` or `drive_id`. I think an alternate way to do it would just be to do `as_dribble()`, check `is_folder()` and then make sure it only identifies 1 row. This eliminates the `get_one_path()` utility, though, and I was worried I'm not thinking of something important that that function does.

R/drive_publish.R
```diff
@@ -78,8 +78,7 @@ drive_change_publish <- function(file = NULL, publish = TRUE, ..., verbose = TRU
     proc_res <- process_response(response)
   }
 
-  ## TO DO: do we really have to call the API again?
-  file_update <- as_dribble(drive_id(file$id))
+  file_update <- drive_id(file$id)
```
jennybc: It still feels like there is at least one gratuitous API call in the typical publishing workflow.

1. Get current metadata.
1. Main API call to enact new publishing status.
1. Optional API call re: auto-publishing.
1. Get current metadata again? This is the one that feels unnecessary to me.

Can you take a look?

R/drive_mkdir.R
jennybc: Yeah I think it should work just like the parent folder in `drive_mv()`. If it's not simple, then it means we need to improve something else. I need to refactor the path stuff in light of all the other changes. Will do shortly.

jennybc: The path changes are done btw.

R/drive_ls.R
```diff
@@ -0,0 +1,33 @@
+#' List Drive files.
+#'
+#' @param path Character. A single path on Google Drive whose contents you
```
jennybc: If this is coming back already, then it also need to accept the folder as ... path, id, or dribble. Like we're doing everywhere else.

tests/testthat/test-drive_ls.R
```diff
@@ -0,0 +1,34 @@
...
+
+nm_ <- nm_fun("-TEST-drive-ls")
+
+run <- FALSE
+clean <- FALSE
+if (run) {
+  ## make sure directory is clean
+  if (clean) {
+    del_pths <- c(nm_("foo"), nm_("this-should-not-exist"))
+    del <- purrr::map(del_pths, drive_delete, verbose = FALSE)
```
jennybc: Isn't `drive_delete()` vectorized? If so, all the deletions in test set up should rely on that, rather than `purrr::map()`.

LucyMcGowan: It is, but I think `drive_path()` isn't :( so I think inputing the character vectors requires the mapping right now.

jennybc: I see. I'm not sure `drive_path()` really can/should take > 1 path. Or maybe it can/should? What do you think? Since we don't include path information, it would seem weird to get files back with no info on which of the input paths they matched.

But when we're using `drive_path()` internally, that seems less of a problem. The mapping could happen inside `drive_delete()`. But then that would apply in many places and we should solve with a function.

R/as_dribble.R
```diff
@@ -34,15 +34,13 @@ as_dribble.dribble <- function(x, ...) x
...
-  drive_path(x)
+  do.call(rbind, purrr::map(x, drive_path))
 }
 
 #' @export
 #' @rdname as_dribble
 as_dribble.drive_id <- function(x, ...) {
-  ## TO DO: we should accept x with length > 1
-  drive_get(x)
+  do.call(rbind, purrr::map(x, drive_get))
```
jennybc: Re:, e.g. `drive_path()` not being vectorized but wanting other functions like `drive_delete()` to accept multiple files. We forgot to circle back to this! This should take care of it. Can you think of downstream changes this suggests? At the very least, some tests of basic dribble formation with vector input would be good.

We probably want to use one of purrr's safer mapping strategies, for when there's one bad path or id among others that work? But perhaps we should open an issue for that and  not let that keep us from merging this.

jennybc:
@LucyMcGowan I cleaned up our `.gitignore`. I'm happy with the effect on my machine, but please make sure it doesn't have unexpected results on yours and lure you into committing any tokens.

Issue #77, opened 2017-06-05
Rename `is_one()`


jennybc:
Not sure what the new name should be but `is_one()` implies it will return logical, which is confusing.

jennybc:
Applies to `is_any()` as well.

Maybe `is_one()` and `is_any()` *should* exist, but live up to their names. We still need the functions that act as valves, only allowing the input to flow through if it has the required number of rows.

LucyMcGowan:
ick I agree `is_one()` and `is_any()` really imply logical output. I wonder if we should have all of our `is_*()` functions, and then have some function that will check it & output a sensible error with glue? Is something like this worthwhile:
```
valve <- function(x, f, msg) {
 if (!f(x)) {
    stop(
      glue::glue_data(x, msg),
      call. = FALSE
    )
  }
}
```

jennybc:
Yeah that's what I had in mind with `msg` in #74. This also seems related to mime type helpers. Other than checking length, a common thing is to make sure you have, say, a folder.

LucyMcGowan:
does the above sketch make sense with 3 inputs? trying to think if there is a more efficient way to do it, but that doesn't seem too bad.

Issue #76, opened 2017-06-03
File identification


jennybc:
We probably want to do some renaming. `drive_from_path()` and `drive_from_id()` better describe what the current `drive_path()` and `drive_get()` functions do. Too clunky though?

It also seems odd that those functions can't take a vector of inputs, but `as_dribble.character()` and `as_dribble.drive_id()` can. Think about this.

All the ways to identify files:

  * `drive_path()`
  * `drive_get()`
  * `drive_search()`
  * `drive_ls()`
  * `as_dribble()` on character vector and `drive_id` vector and (for internal use) list of Files Resource objects or `dribble`

The ones that start with `drive_` are conceived to be called directly by a user. `as_dribble()` is generally used internally to pre-process user input describing the file(s) to act on.

LucyMcGowan:
would it be very strange if `as_dribble()` was the function we encouraged the user to use? 

*pros*: it could  be a bit simpler - this is basically converting whatever you have (path, id, list) to a `dribble` that we can recognize with some basic checks. Also, it works on vectors, which seems natural. The workflow would basically be I have some "thing" & I need to register it as a `dribble` (but if I use a drive function explicitly, it will be registered automatically).

*cons*: maybe it's confusing because the user isn't explicit about what they think they're inputting? Also if they wanted to input an id they'd have to use `drive_id() %>% as_dribble()`.

jennybc:
> would it be very strange if `as_dribble()` was the function we encouraged the user to use?

Maybe not. I feel like we're still converging on the most natural way for people to get a hold on the files they want to work on. The good news is that all of it is now possible and fairly smooth.

Issue #75, opened 2017-06-03
safer purrr mapping for drive_ids


LucyMcGowan:
* We probably want to use one of purrr's safer mapping strategies, for when there's one bad path or id among others that work?

jennybc:
The non-existence of a file specified by id throws an error. If specified by path and no file matches, you just don't get anything.

jennybc:
I'm not convinced there's a clear action to take here. Perhaps wider usage will reveal if we need to do anything on this front.

PR #78, opened 2017-06-05
Vectorize drive_get() and rationalize drive_path*(); closes to #76

jennybc:
I see no reason why `drive_get()` should not be vectorized. When n things go in, n things come out (we still need to think about bad ids #75).

The functions that take path as input are trickier: there's the issue of whether the input is allowed to have length > 1 (which I have straightened out here) and whether the output is allowed to have length > 1 (which is a job for whatever `is_one()` becomes #77).

I still want to add some tests, but overall @LucyMcGowan do you see any problem with this? Also, can you think of downstream adjustments we can or need to make?


LucyMcGowan:
I think this looks 👍 - the only thing I think is a bit tricky is figuring out how to tell someone if *some* of the ids are bad (or maybe it's all or nothing - if one is bad it stops regardless?)

jennybc:
Let's leave the "bad ids" problem in #75. Not sure if it should error or those rows should be non-existent or filled with NAs.

I think there are downstream implications here:

  * Anything that was counting on `drive_path()` to make sure there was only 1 input needs to switch to `is_one()`.
  * Anything that could benefit from vectorization, such as `drive_delete()`, should do it internally now.

LucyMcGowan:
are you working on those 👆, or should I take a stab at them?

jennybc:
I propose you tackle the second (search for functions that can clearly be vectorized but aren't) and I'll work on the first.

I think you will find many functions that *could* be vectorized but perhaps we should not rush into it, such as moving files or modifying publishing/sharing status. Recommend you start with the ones that are no-brainers. It might just be `drive_delete()`? As you encounter ones that require more thought, open an issue.

LucyMcGowan:
I think the obvious ones are `drive_delete()`, `drive_share_link()`, and `drive_is_published()` - I believe each of those can currently handle `dribble`s with multiple rows, so their code wouldn't change with this update since `as_dribble()` would just output an n-rowed `dribble` versus a 1 row dribble.

LucyMcGowan:
😱 `as_dribble.NULL` is brilliant!

PR #80, opened 2017-06-08
Incorporate discovery document (fixes #29)

### LINKED ISSUE #29: Re-wrangle stored info re: Drive API
From #20 
> You might want to consider a two-layer request builder, like here:
> https://github.com/tidyverse/googlesheets2/blob/8caf705a92b7c0638783ce478eb0449a3a92c1de/R/gs_build_request.R#L47

> The inner layer, gs_build_request(), is very dumb and mechanical. But would work if someone really knew what they were doing and only wanted help setting base URL, path param expansion, and managing auth.

> The outer layer, gs_generate_request(), is smarter and what I plan to use! It needs to be given a known endpoint and then it looks up everything we know, checks that input looks good, and massages it into shape before calling gs_build_request().

LucyMcGowan:


LucyMcGowan:
@jennybc I think this is ready for a little 👀. I based it mostly on how googlesheets4 works, but I ended up having to remove the interior `build_request()` - since I needed to parse out body and query parameters, I would have had to pass the endpoint to the interior function, so it seemed kind of unnecessary since it would have been almost exactly the same as the outer function (if you think there is a better way to do that, I'm open to it though!). Also thank you for helping to bring in the schema parameters - your way is so much more elegant 💃

R/drive_upload.R
```diff
@@ -105,16 +105,26 @@ drive_upload <- function(from = NULL,
...
-    endpoint = "drive.files.update.media",
-    params = list(
-      fileId = up_id,
-      uploadType = "media",
-      body = httr::upload_file(path = from,
-                               type = mimeType),
-      fields = "*"
-    )
-  )
+  request <- list(url = httr::modify_url(url = .drive$base_url,
```
LucyMcGowan: this is not ideal -- the path for updating a new file is a bit buried in the discovery document, so I haven't pulled it out yet, but I should update that soon. @jennybc the drive.files.update endpoint is a bit tricky - in our previous iteration, I manually created two separate endpoints: drive.files.update.meta and drive.files.update.media. Do you think we should do that again, or is there a more elegant way to handle this?

jennybc: It sounds like a good idea to do what you need to do over in the discovery document prep, i.e. to create two separate entries in our `endpoints` object. It feels cleaner to do special handling there vs. in these functions.

R/drive_user.R
```diff
@@ -38,10 +38,10 @@ guser <- function(fields = "user", ...) {
...
 
-  request <- build_request(
-    endpoint = "drive.about.get",
-    params = list(fields = fields,
-                  ...))
+  request <-  list(url = httr::modify_url(.drive$base_url, path = paste0("drive/v3/about")),
+                   query = list(fields = fields),
+                   method = "GET",
+                   token = drive_token())
   response <- make_request(request)
```
jennybc: Shouldn't this also be one of our endpoints? Seem like we could do same for "About" as we're doing for files, permission, and revisions.

R/drive_upload.R
LucyMcGowan: working on this now 👆

data-raw/mime-types.R
```diff
@@ -13,8 +13,10 @@ google_mime_types <- httr::GET(url) %>%
   select(mime_type = `MIME Type`)
 
 
-fmts <- build_request(endpoint = "drive.about.get",
-                      params = list(fields = "importFormats,exportFormats")) %>%
+fmts <- list(url = httr::modify_url(.drive$base_url, path = paste0("drive/v3/about")),
+             query = list(fields = "importFormats,exportFormats"),
+             method = "GET",
+             token = drive_token()) %>%
```
jennybc: Now that we have the About endpoint, can we use it here?

LucyMcGowan: I forgot to update that one 🙈 - will do it now.

R/drive_request.R
```diff
@@ -44,20 +44,30 @@ build_request <- function(endpoint = character(),
...
+
+  out$method <- ept$method
+
+  if (grepl("media", endpoint)) {
+    out$path <- glue::glue_data(params$path_params, ept$path)
+  } else {
+    out$path <-  paste0("drive/v3/", glue::glue_data(params$path_params, ept$path))
+    out$body <- c(params$body_params)
+  }
+  out$query = c(params$query_params)
```
jennybc: This feels too special case-y. What exactly is different for the media endpoints? AFAICT from reading this:

  * The `drive/v3` prefix for the path needs to be suppressed.
  * Something about preparing the body?

I hope we can find a cleaner way to accommodate them.

LucyMcGowan: I agree! Yes, the two things are as you mentioned. The body for media uploads seems to need to be unnamed and contain exactly the output from `httr::upload_file()`.

jennybc: As for the `drive/v3` prefix: Looking at our `.endpoints` object, why don't you prefix the `path` in the prep of `.endpoints` for everything except the two media endpoints? That is, fully prepare `path` in the pre-wrangling.

jennybc: As for the body, could you handle that in the small number of functions that upload media? So call `build_request()` in the usual way but edit the body component prior to `make_request()`? I would just like to get all this special logic out of the basic request builder.

jennybc:
This looks ready to merge, yes?

LucyMcGowan:
I think so, yes 🎉 I should squash?

jennybc:
Actually, you need to re-`document()`.

LucyMcGowan:
Do I? I just ran it and it didn't change anything. Or do you mean I need to change the write-up on the build functions?

jennybc:
Sure, I think you should squash and merge now. Do you have "fixes #29" in one of the commit messages? You might want to put that in the overall title, in fact.

LucyMcGowan:
I do, but I'll stick it in the title too 🎉

PR #82, opened 2017-06-09
split build_request into 2 functions

LucyMcGowan:
I've 
* split `build_request()` into `generate_request()` with an inner `build_request()` function a la googlesheets4
* slightly updated docs to reflect this change


jennybc:
If my additions look OK to you @LucyMcGowan, then go ahead and merge this.

PR #83, opened 2017-06-11
Provide some helpers for field handling; fixes #39

### LINKED ISSUE #39: Fields handling
Moved out of #7.

~~We need to rationalize `fields` and the columns that appear in the result. Do we equate these two? This feels connected with writing a general handler for an instance of [file resource](https://developers.google.com/drive/v3/reference/files) (Drive-speak).~~

jennybc:
This is still a WIP, but am interested in feedback.

Do you remember @LucyMcGowan how you can tell whether an endpoint needs the fields to have `files/` prepended? Obviously I know other endpoints have fields, but we're really focused on the ones relevant to a Files resource. It would be nice to do something intelligent with that info.

Also, wasn't there a shortcut for essentially specifying the `files/` prefix for a set of several fields?

Any other thoughts? This is for internal use and for us or others to use in more "manual" settings. I've been playing with `build_request()` calls and have used it successfully.

See #39


LucyMcGowan:
I haven't found a doc that explains this fully, but I've mapped a bit of a pattern. When the resource is grabbing a specific thing (like with `drive.files.get`) we can use the `field` parameter (for example we can say `fields = id`, but if we are listing multiple things (like with `drive.files.list`) we have to specify that the subgroup we want to get all the fields for is `file` so we specify `fields = files/id`. Similarly, using the permissions resource, if we want to get the permission id if we are using `drive.permissions.get` we can just say `fields = "id"`, but if we are using `drive.permissions.list` we have to say `fields = "permissions/id"`. 

There is a brief explanation [here](https://developers.google.com/drive/v3/web/performance#partial), but it doesn't map all of them.

PR #86, opened 2017-06-12
update as_id() to accommodate Google Drive URLs

### LINKED ISSUE #85: extract ids from urls
* write a small function to extract `fileId`s from the urls
* Notes:
    *  most urls have the form: `https://docs.google.com/*/d/fileId/edit` (where * is document, spreadsheets, forms, or presentation)
    * in all of the links that I have found so the`fileId` is preceded by /d/
    * found this [SO question](https://stackoverflow.com/questions/16840038/easiest-way-to-get-file-id-from-url-on-google-apps-script) that verifies that this is mostly true
    * Craig's advice: stick with /d/ - if the user has links that don't follow this pattern have them open in the browser and copy drive's redirect

LucyMcGowan:
This is a very simple function to extract the `fileId` from a url. It relies on the url to have /d/ before the fileId. If there is not a `/d/` in the url, it returns an NA.


LucyMcGowan:
@jennybc I've made a very basic function (`drive_extract_id()`) that will simply grab the id after `/d/` and then I created a second function that will certainly need a different name (`drive_extract_id_smarter()`) that will first try to find the id after `/d/`, but if `/d/` doesn't exist, it will try the url to see if it is an old one that has a proper redirect to `/d/`. Let me know if you think this is proper!

LucyMcGowan:
@jennybc since it was such a little function I ended up adding the mime type function as well.

R/utils.R
```diff
@@ -44,3 +44,13 @@ redirect_id <- function(id, url) {
...
+
+get_mime_type <- function(x) {
+  if (!(x %in% c("document", "spreadsheet", "folder", "presentation",
+                 "form", "drawing", "script"))) {
+    message(glue::glue("Input was not a Google Drive type: {x}"))
+    invisible(NULL)
+  } else {
+  paste0("application/vnd.google-apps.", x)
+  }
+}
```
jennybc: Let's split this into a separate PR? I think it should be doing a lookup, just on general principle, but I confess I can't talk you why we can't do the above for now. Except, for example, I think if you put a real mime type in, it should pass through unharmed.

Again, let's do this in a different branch/PR.

jennybc:
My main thought: I think your "id extraction function" should be `as_id()` and you can rename / merge with `drive_id()`. So alleged ids go in and validated ids (in some weak sense -- not by calling the API), with the correct S3 `drive_id` class, come out. The regex (or regexes?) should be able to recognize a naked id and pass it through or use the logic you've got to detect the typical URLs and extract the id.

R/utils.R
jennybc: I think basically the current `ext` column in `.drive$mime_tbl` should also be valid input. As in `get_mime_type("pdf")` returns `application/pdf`.

jennybc:
Also, this will be an exported function, so I think it deserves it's own file and documentation.

R/as_id.R
```diff
@@ -14,21 +14,19 @@ as_id <- function(x) {
...
+    return(x)
+  }
+  ## We expect the links to have /d/ before the file id, have /folders/
+  ## before a folder id, or have id= before an uploaded blob
+  id_loc <- regexpr("/d/([^/])+|/folders/([^/])+|id=([^/])+", x)
+  if (id_loc == -1) {
+      NA_character_
+  } else {
+      gsub("/d/|/folders/|id=", "", regmatches(x, id_loc))
   }
```
jennybc: This is a good place to do an early return and have less nesting of the main function body.

tests/testthat/test-as_id.R
```diff
@@ -1,31 +1,34 @@
 context("Drive id class")
 
-test_that("as_id() properly returns character strings as ids", {
+test_that("as_id() returns non-URL character strings as ids", {
 
   expect_s3_class(as_id("12345"), "drive_id")
 
-  expect_equal(as_id("12345")[1], "12345")
+  expect_identical(as_id("12345")[1], "12345")
```
jennybc: We can use `expect_identical()` here, so I did.

jennybc:
🎉

PR #90, opened 2017-06-13
Mime type lookup; fixes #52

### LINKED ISSUE #52: MIME type helpers and checkers
* condense into a lookup function
* should be used `drive_upload()` and `drive_list()`
* function where user gives input (file) and it will tell you what mime type(s) it can be uploaded as
* sketch:
```
drive_mimetype(name = , type = , mime_type = )
```
`name` would be the name of a file to check, `type` would human readable type, `mime_type` would be the full mime type. You'd just fill in one.

other notes:
* `drive_upload()` shouldn't default to folder
* "Google Doc" "Google Sheet" "Google Slide"
* in `drive.mime_type`
    * create column "type" variable for the ones without extensions (generally the Google types) for easier lookup
    * create column "group" (maybe later)

LucyMcGowan:


LucyMcGowan:
@jennybc does this look 👌 to merge?

R/drive_mime_type.R
```diff
@@ -0,0 +1,33 @@
+#' Lookup a mime type from a Google Drive type or file extension
+#'
+#' @param type Character. The Google Drive type or file extension you would
+#'   like to look up the mime type of.
```
jennybc: I think this should be the definitive place to document the `type` argument. And other functions should use `@inherit` or `@template`, so you don't repeat it everywhere.

R/drive_mime_type.R
```diff
@@ -0,0 +1,33 @@
...
+#' @examples
+#' ## get the mime type for Google Spreadsheets
+#' drive_mime_type("spreadsheet")
+#'
+#' ## get the mime type for jpegs
+#' drive_mime_type("jpeg")
+#' @export
+drive_mime_type <- function(type = NULL, verbose = TRUE) {
+
+  if (!(is.character(type) && length(type) == 1)) {
```
jennybc: Is it really necessary to limit this to `type` of length 1? Seems like it would be just as easy to make this vectorized.

R/drive_mime_type.R
```diff
@@ -0,0 +1,33 @@
...
+    stop("Please update `type` to be a character string.", call. = FALSE)
+  }
+
+  mime_type <- .drive$mime_tbl$mime_type[.drive$mime_tbl$mime_type == type |
+                                           (!is.na(.drive$mime_tbl$human_type) &
+                                              (.drive$mime_tbl$human_type == type))
+                                         ]
```
jennybc: I believe this should be based on `match()`.

R/drive_mime_type.R
```diff
@@ -0,0 +1,33 @@
...
+#'
+#' @examples
+#' ## get the mime type for Google Spreadsheets
+#' drive_mime_type("spreadsheet")
+#'
+#' ## get the mime type for jpegs
+#' drive_mime_type("jpeg")
+#' @export
+drive_mime_type <- function(type = NULL, verbose = TRUE) {
+
```
jennybc: I think this should return the MIME type table when given no input.

R/drive_search.R
```diff
@@ -47,6 +52,13 @@ drive_search <- function(pattern = NULL, ..., verbose = TRUE) {
     params$fields <- paste0("files/", .drive$default_fields, collapse = ",")
   }
 
+  if (!is.null(type)) {
+    mime_type <- drive_mime_type(type, verbose = verbose)
```
jennybc: You can use the `x <- x %||% something_else()` pattern here.

R/drive_upload.R
```diff
@@ -45,15 +45,10 @@ drive_upload <- function(from = NULL,
...
   name <- name %||% basename(from)
 
-  ## mimeType
-  if (!is.null(type) &&
-      type %in% c("document", "spreadsheet", "presentation", "folder")) {
-    type <- paste0("application/vnd.google-apps.", type)
+  mimeType <- NULL
+  if (!is.null(type)) {
+    mimeType <- drive_mime_type(type)
   }
```
jennybc: You can use the `x <- x %||% something_else()` pattern here.

R/drive_mime_type.R
LucyMcGowan: I originally had it vectorized and then switched it 🙈

R/drive_mime_type.R
```diff
@@ -0,0 +1,33 @@
...
+  mime_type <- .drive$mime_tbl$mime_type[.drive$mime_tbl$mime_type == type |
+                                           (!is.na(.drive$mime_tbl$human_type) &
+                                              (.drive$mime_tbl$human_type == type))
+                                         ]
+  if (length(mime_type) == 0L) {
+    if (verbose) {
+      message(glue::glue("Ignoring `type` input. We do not have a mime type for files of type: {sq(type)}"))
+    }
+    return(invisible(NULL))
```
jennybc: Maybe it's correct to error if any input can't be found? Probably so. But we could also consider "n in, n out" and put `NA`s or pass the input through, for non-matches.

R/drive_search.R
jennybc: Oh wait, no you can't! This is the other way around, where you want to preserve `NULL`.

R/drive_mime_type.R
LucyMcGowan: maybe an error or NAs - the way it's set up now I don't think we want to pass the input through because I use it in `drive_upload()` and `drive_search()` as the set `mimeType`, so we want to make sure it outputs an actual mime type or nothing?

R/drive_search.R
```diff
@@ -47,6 +52,13 @@ drive_search <- function(pattern = NULL, ..., verbose = TRUE) {
...
+    if (!is.null(mime_type)) {
+    params$q <- glue::collapse(c(params$q, paste0("mimeType = '", mime_type,"'")),
+                               sep = " and ")
+    }
+  }
```
jennybc: By picking the right functions, you can probably make this unconditional on whether `mime_type = NULL`. I.e., I think you could use `glue::glue()` instead of the `paste0()` and it will return `character()` if `mime_type = NULL`. Then probably `paste()` instead of `glue::collapse()` to combine `q` and formatted MIME type.

R/drive_upload.R
jennybc: Ditto my retraction from above!

R/drive_mime_type.R
jennybc: I think the property that a real MIME type passes through unchanged is attractive. It's generally good if you could pass the output of `drive_mime_type()` through it again and not change anything.

But that doesn't mean you have to pass *anything* through. For input that doesn't match the human types, you could attempt a match against actual MIME types.

It's possible you should have another little helper that catches `NA`. Or you could let the API raise the error about a bad MIME type.

R/drive_mime_type.R
LucyMcGowan: ⚠️ This is a silly question: How would I do it with `match()`? I think I want to output the mime type that is equal to the human type input rather than checking if it is `%in%` the list? I am very likely missing something though!

jennybc: Sorry, you will index `.drive$mime_tbl$mime_type` by the result from `match()`.

R/drive_mime_type.R
LucyMcGowan: I think what you are suggesting is what it does now - you can pass either the `human_type` or the `mime_type` and it will output the `mime_type`- if it matches neither, it will output NULL (but this will change to NA when I vectorize it).

R/drive_mime_type.R
jennybc: Along these lines:

```r
type <- c("jpg", "folder")
.drive$mime_tbl$mime_type[match(type, .drive$mime_tbl$human_type)]
```

R/drive_mime_type.R
jennybc: OK, yes then, sounds good.

R/drive_mime_type.R
LucyMcGowan: Ah and then if it's NA check if it's in `mime_type`?
```
type <- c("jpg", "folder")
mime_type <- .drive$mime_tbl$mime_type[match(type, .drive$mime_tbl$human_type)]
if (is.na(mime_type)) {
mime_type <- .drive$mime_tbl$mime_type[match(type, .drive$mime_tbl$mime_type)]
}
```
The way I have it now it first checks if it's just a mime_type and if not it checks if it's a human type.

LucyMcGowan: But then if you passed some human types and some mime types it wouldn't know which to replace 🤔

jennybc: Maybe more like this:

```r
type <- c(NA, "folder")
m <- match(
  type,
  .drive$mime_tbl$human_type,
  nomatch = NA_character_,
  incomparables = NA
)
.drive$mime_tbl$mime_type[m]
```

LucyMcGowan: but don't we want you to be able to pass "application/vnd.google-apps.folder" and get "application/vnd.google-apps.folder"?

jennybc: Sorry, maybe I misunderstood what you were asking? I thought you were asking about `NA`s on the way in.

LucyMcGowan: Oh no I was thinking of a way to first check if it's in `human_types` and if not, check if it's in `mime_types` and if neither give NA. So I did that previously like this: 
```
.drive$mime_tbl$mime_type == type |
  (!is.na(.drive$mime_tbl$human_type) &
  (.drive$mime_tbl$human_type == type))
```
But I'm not sure how to use `match()` such that it can check either?

R/drive_mime_type.R
```diff
@@ -13,21 +12,33 @@
...
-  mime_type <- .drive$mime_tbl$mime_type[.drive$mime_tbl$mime_type == type |
-                                           (!is.na(.drive$mime_tbl$human_type) &
-                                              (.drive$mime_tbl$human_type == type))
-                                         ]
-  if (length(mime_type) == 0L) {
-    if (verbose) {
-      message(glue::glue("Ignoring `type` input. We do not have a mime type for files of type: {sq(type)}"))
-    }
-    return(invisible(NULL))
+  if (all(is.na(mime_type))) {
```
jennybc: Nice: this ends up being a good solution where we get to vectorize this, but can still use it easily inside functions that work on a dribble holding one file.

R/drive_search.R
```diff
@@ -53,11 +53,13 @@ drive_search <- function(pattern = NULL, type = NULL, ..., verbose = TRUE) {
...
-    if (!is.null(mime_type)) {
-    params$q <- glue::collapse(c(params$q, paste0("mimeType = '", mime_type,"'")),
-                               sep = " and ")
-    }
+    mime_type <- drive_mime_type(type)
+    ## if they are all NA, this will error, because drive_mime_type
+    ## doesn't allow it, but if one is NA and the others aren't,
+    ## this will still pass the ok ones through.
+    params$q <- paste(c(params$q,
+                        paste0("mimeType = '", mime_type,"'", collapse = " or ")),
```
jennybc: Hmmm, I hadn't thought about multiple MIME types here. I think we should probably only allow one? Because with `q`, we are going to collapse multiple clauses with "and", so it feels weird to combine multiple `type`s with "or". If people want an "or" clause for mime types, they can build it themselves and provide via `q`.

jennybc: I've left this (I'm about to push), but did tweak so we drop the `NA`s before forming the `q` clause. Maybe this doesn't hurt anything: `"mimeType = 'text/csv' or mimeType = 'NA' or mimeType = 'NA' or mimeType = 'application/pdf'"` but it seems odd.

Issue #81, opened 2017-06-08
drive_read()?


tiernanmartin:
Is there a way to read a data object out of a `dribble` (similar to `readr::read()` or `googlesheets::gs_read()`)?

jennybc:
Not yet but it could happen. Question is: what sort of object do you read it into? Sheets are one thing and will be handled by googlesheets. But what else might you want to read straight into an R object? 🤔 Curious to hear the use you have in mind.

tiernanmartin:
While I do use Google Drive to create and edit spreadsheets, I also use it as cloud-based data storage for file formats that aren't editable through the Drive interface. These file formats include `.csv`, `.md`, `.rds`, `.zip`, `.gpkg`, etc.

So I guess I'd like a function that could serve a wrapper around the `read_*` functions, giving them access to my Drive account.

Something like:

```r
## this code is for illustration purpose - not a functional reprex

# Upload file to Drive
write.csv(chickwts, "chickwts.csv")
drive_chickwts <- drive_upload("chickwts.csv")

# Read that file back into R
drive_read(obj = drive_chickwts , fun = read_csv, fun_args = list()) # fun_args would be passed to the read function's params
```

jennybc:
It's been a while but if you still care about this @tiernanmartin, I'd love to hear if it meets your needs.

tiernanmartin:
Nice! This looks great.

I am not sure whether it will meet my particular needs. I work with a lot of geopackage (.gpkg) files that I store on google drive, and my naïve attempt to access one of those with this new function returned an error. 

Here's a reprex:

``` r
library(tidyverse)
library(sf)
library(googledrive)

nc <- st_read(system.file("shape/nc.shp", package="sf"))

st_write(nc, "nc.gpkg")

nc_in_drive <- drive_upload("nc.gpkg")

drive_read(nc_in_drive) %>% 
  st_read()

## Error in CPL_read_ogr(dsn, layer, query, as.character(options), quiet,  : 
## Expecting a single value: [extent=167936].
## In addition: Warning message:
## In if (nchar(dsn) < 1) { :
## the condition has length > 1 and only the first element will be used
```

jennybc:
I'm looking at the docs for `sf::st_read()`:

https://r-spatial.github.io/sf/reference/st_read.html

The first argument `dsn ` is documented as:

[...]

Issue #44, opened 2017-05-22
Uploading an R object w/o writing to file


jennybc:
In testing, I note its awkward to always have to write to file before uploading. We need to offer a way to upload from an R object directly to Drive.

sanjmeh:
a gentle reminder that this is still open.
One more reason why writing to a file first is a bad idea is because the NAs are all lost. Also sometimes extra quotes are inserted that need cleaning. Thanks

jennybc:
I can't convince myself this is needed. To implement, you have to have specific combinations of R object and destination file type in mind. The most common one would be a data frame and Sheet, but that's what googlesheets4 is for. Closing until a specific use case presents itself.

jennybc:
I can't convince myself this is needed. To implement, you have to have specific combinations of R object and destination file type in mind. The most common one would be a data frame and Sheet, but that's what googlesheets4 is for. Closing until a specific use case presents itself.

renatocava:
> I can't convince myself this is needed. To implement, you have to have specific combinations of R object and destination file type in mind. The most common one would be a data frame and Sheet, but that's what googlesheets4 is for. Closing until a specific use case presents itself.

Hello @jennybc, a necessity is when you want to use Google Drive to store backup copies that are created daily, then an RDS or CSV file will take up less space than a GOOGLESHEET. Any recommendation?

jennybc:
@renatocava In general, closed issues aren't a good place for discussion. If you're really interested in this, you should open a new issue and, in this case, provide some concrete details about the size differences you observe.

PR #91, opened 2017-06-13
build drive_download()

LucyMcGowan:
This is a WIP - working out how to best download files from Google Drive.

@jennybc I've changed the philosophy a bit from the first push. Now the user can download a single file using `drive_download()`, which will output an object of class `drive_file` containing the raw data from Google Drive, parsed by httr. 

Note: Right now the code is pretty light because I rely on [httr::content() auto-parsing](https://github.com/r-lib/httr/blob/1fc659856602f60ff75eb01903513244e3491ec2/R/content-parse.r) (which I know is a no-no 😬) - I've linked to how they elect to parse from mime type - I think we should do something similar, but don't want to reinvent the 🎡?

Here is how it would work:
```
chickens <- drive_download("chickwts.csv")
```
Here `chickens` will be a `tibble` with the contents of chickwts.csv.

This will extract the .csv and save it locally to the working directory as "chickens.csv":
```
drive_read_csv("chickwts.csv", path = "chickens.csv")
```


jennybc:
Hi @LucyMcGowan, I'm about to be in transit for ~24 hours. But just to check in here ... as we discussed on Slack, I think this should handle the simple and common Drive file --> local file case. I would expect `drive_download("chickwts.csv")` to leave `chickwts.csv` as a local file in current working directory. I would expect to be able to optionally specify the local name/path.

As for Drive file --> R object in memory ... I still think that's worth doing, but shouldn't clutter up the simple `drive_download()` workflow. It is good to use common functions underneath, so good to plan to this. That will soon bring us to `drive_read()` (#81) and `drive_write()` (#44).

LucyMcGowan:
`drive_download()` could definitely do that but I think it is a little more complicated than how we have it now because (unless there is something obvious I've missed) we have to specify what to write with depending on the output file, so we'd have to pull the file in as an R object either way and then choose the writer (for example `write.csv()` for .csvs or `writeBin()` for .rda files) but you're probably correct that all that should be in `drive_download()`.

jennybc:
I think you can use `httr::write_disk()` to download to disk. I just used used this snippet to successfully download a chicken small png (I'm authenticated using our shared tidyverse account).

```r
x <- drive_get("0B0Gh-SuuA2nTeDQzRUFXWU5hRGs")
download_link <- x$files_resource[[1]]$webContentLink
res <- httr::GET(
  download_link,
  httr::write_disk("chicken_small.png"), drive_token()
)
```

Do you think you can move forward with this basic idea?

jennybc:
There may be a better or more proper way, using the file's resource URL and `alt=media` (now I'm just reading the [download guide](https://developers.google.com/drive/v3/web/manage-downloads), which I assume you are too). But I think you should be able to write to directly to disk.

For Google native docs, you'll need to export. But I guess there could be a default for each flavour and user could optionally specify the export type.

LucyMcGowan:
Ohhhh I wasn't using `httr::write_disk()` yes I can do that 👌🏼 Yes right now the function is using the `alt=media` URL, I assume that works with `write_disk()` as well -- I'll switch to this.

jennybc:
So here's my example, re-phrased a bit more in our function-vocabulary:

```r
req <- generate_request(
  "drive.files.get",
  params = list(fileId = "0B0Gh-SuuA2nTeDQzRUFXWU5hRGs",
                alt = "media")
)
resp <- make_request(req, httr::write_disk("chicken_small.png"))
```

jennybc:
I think this last example is closer to what one should be doing, to be clear.

LucyMcGowan:
@jennybc 👌 I think we're on the same page now (I've gone in circles a bit on this one, so definitely look at the overall changes versus any from a single commit to avoid confusion). Thank you!

R/drive_download.R
```diff
@@ -1,60 +1,84 @@
...
+                           out_path = NULL,
+                           type = NULL,
+                           overwrite = FALSE,
+                           verbose = TRUE) {
   file <- as_dribble(file)
   file <- confirm_single_file(file)
 
+  if (is.null(out_path)) {
+    out_path <- file$name
+  }
```
jennybc: `out_path <- out_path %||% file$name`

R/drive_download.R
```diff
@@ -1,60 +1,84 @@
...
-  response <- make_request(request)
-  httr::stop_for_status(response)
-  proc_res <- httr::content(response, encoding = "UTF-8")
+  } else {
+    ## TODO make sure the type is compatable.
+    type <- type %||% tools::file_ext(out_path)
+    if (type == "") {
+      ## TODO, be able to guess type by default (for example saving sheets
+      ## as csvs..)
+      stop("We cannot guess type yet")
```
jennybc: So your plan is to use your knowledge of `mime_type` and pick a default export type here? Sounds like a good plan. Do you want to go ahead and do that while we're at this? FWIW I would consider defaulting to Excel for Sheets, because I think csv will only get you the first sheet.

This sounds like another mime type helper -- `drive_mime_export()`? -- that probably works like so:

  * `drive_mime_export()` with no input returns a tibble with ... 4 variables? Basically the "export" rows of `.drive$translate_mime_types`. It seems like we need a `default` variable where you can designate a default, for any MIME type with more than one export type (maybe that is only the Google types?). Which you would then use right here.
  * `drive_mime_export(x)` returns the default export MIME type for the input MIME type
  * `drive_mime_export(x, as = y)` checks that something of MIME type x can be exported as MIME type y. Is this useful / needed?

I can't think of any reason to NOT vectorize this on x. Is it going to need a companion for import? If so, I'd consider that too. At least make sure that the set of MIME type helper functions make sense (or will) as a group.

R/drive_download.R
```diff
@@ -1,60 +1,84 @@
...
   file <- confirm_single_file(file)
 
+  if (is.null(out_path)) {
+    out_path <- file$name
+  }
+
+  mime_type <- file$files_resource[[1]]$mimeType
+
+  if (!grepl("google", mime_type)) {
+
```
jennybc: It feels like you can simplify the logic around here, i.e. have less code down inside if () {...} else {...}. Something like

```r
params <- list(file = file$id, alt = "media")
if (!grepl("google", mime_type)) {
  #adjust params for this case
}
request <- generate_request(endpoint = "drive.files.export", params = params)
```

R/drive_download.R
```diff
@@ -1,60 +1,84 @@
...
+                                params = list(fileId = file$id,
+                                              mimeType = mime_type ))
+  }
+
+  ## it seems that it needs an extenstion to save properly
+  ext <- tools::file_ext(out_path)
+  if (ext == "") {
+    ext <- .drive$mime_tbl$ext[.drive$mime_tbl$mime_type == mime_type]
+    out_path <- paste0(out_path, ".", ext)
+  }
```
jennybc: Maybe do this at the top, when you're already working on `out_path` the first time?

jennybc:
This looks really close. I made a couple of comments, then it needs some tests and inclusion in the README. It's also going to sign you up for some MIME type helper work.

jennybc:
It seems I was commenting on a commit, vs cumulative changes, so you might need to read my comments with that in mind. Oops.

jl5000:
I've been attempting to use this package to read csv files stored on Googledrive directly into R dataframes (I've already downloaded them onto my local folder but I'm looking for something a bit neater without the need to download and store). I can't really tell from the above discussion but is this something that's achievable now or on the horizon with this package?

jennybc:
I think this is #81, which you are welcome to chime in on.

PR #95, opened 2017-06-14
update readme with file creation/deletion (closes #92)

### LINKED ISSUE #92: Setup and teardown for README and each article
We need at least the same low tech solution we are using for the tests: well-preserved code for setup and teardown. I just failed to re-build the pkgdown site because:

```
Updating README.md
Error in drive_upload("chickwts.csv") : Path already exists:
My Drive/chickwts.csvPath already exists:
0AEGh-SuuA2nTUk9PVA/chickwts.csv
```

I deleted it manually, succeeded with site rebuild, and moved on. But we need to make that even easier.

LucyMcGowan:
@jennybc is this how you imagined it? Just wanted to double check before I merge.

### DIFF
--- a/R/as_id.R
+++ b/R/as_id.R
@@ -1,4 +1,4 @@
-#' File id from string or URL
+#' File id from string or URL.
 #'
 #' Marks a character vector as holding Google Drive file ids, as opposed to file
 #' names or paths. User can supply either a Google Drive id or a Google Drive URL.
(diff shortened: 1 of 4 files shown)


jennybc:
Yeah pretty much! But in this case, should that initial chunk be non-conditional? So always check if that file exists and, if it does, delete it? I don't think we need the extra layer of control here.

LucyMcGowan:
oh that's true! I'll update that & add the `as_id()` documentation.

jennybc:
I added more to my first comment FYI.

jennybc:
LGTM other than my small suggestion re: title.

PR #98, opened 2017-06-15
add tests for drive_mkdir() and drive_upload() folder logic

LucyMcGowan:


tests/testthat/test-drive_mkdir.R
```diff
@@ -0,0 +1,29 @@
...
+
+  foo <- drive_path(nm_("foo"))
+  ## foo
+  ## |- bar
+  bar <- drive_mkdir(nm_("bar"), path = nm_("foo"))
+  expect_identical(bar$files_resource[[1]]$parents[[1]], foo$id)
+
+  ## clean up
+  drive_delete(nm_("bar"))
+})
```
jennybc: For this one, maybe the test should just make `foo` then make `bar` within it. So, don't worry about setup other than deleting any conflicting pre-existing files? Seem legit here and makes life simpler.

LucyMcGowan: 👍 so are you proposing we create and delete `foo` & `bar` every time?

jennybc: Yes. It's possible you can just delete the parent folder, but the whole folder situation is so weird that I would check to be sure.

LucyMcGowan: yes! Sorry I deleted my previous comment 🙈 - the last commit does just that 👆. The test itself just deletes the parent folder, but the clean up checks both just in case something funky happens.

PR #103, opened 2017-06-22
Gargle creds; closes #55, closes #56

### LINKED ISSUE #55: API key
We should send an API key with every request, because there are fruitful things that can be done w/o auth, such as accessing world-readable files.

@craigcitro is going to tell us about a specific header to send that clarifies whether the API key or auth token should be consulted re: quota, in the case where both are present in the request.

### LINKED ISSUE #56: Tidyverse-wide project
Create a project to own API key and client id and secret that we will use uniformly across googledrive, googlesheets*, and bigrquery.

jennybc:
Start using the tidyverse-wide Google credentials (#56). They will eventually be obtained from gargle (#102), but this gets the ball rolling.

Start sending an API key in every request (#55).

Also: give much more detailed error messages. I spent a lot of time looking at HTTP errors on this PR!

ATTENTION @LucyMcGowan !!!

Any pre-existing tokens you have will need to be deleted and remade! This PR changes the "app", i.e. the client id and secret. So any pre-existing token cannot be refreshed with the new app.

I have passed tests locally, with this PR, but **only by remaking the testing token**.


jennybc:
I'm going to close this, as I've now pulled the semi-unrelated bits into commits on master. Will open a new PR that's focused on the credentials switch and the addition of API key.

PR #110, opened 2017-06-25
Pagination; closes #87, closes #89

### LINKED ISSUE #87: Pagination
We need to traverse pages for certain endpoints, such as the files list one called by `drive_search()`..

### LINKED ISSUE #89: drive_search() could have an n argument
* it would be nice to only pull n from the search
* be able to traverse pages 
* potentially expose `pageSize` better in the docs

jennybc:
I'll leave this open for a day or two in case you @LucyMcGowan have any feedback.

Closes #87 Pagination. It doesn't handle every last type of pagination in the Drive API but handles the type used by the endpoint we care about: files. Happily, that also seems to cover the comments and, I think?, the revisions endpoints as well.

Closes #89 re: allowing user to cap the number of items retrieved by `drive_search()`.


LucyMcGowan:
@jennybc this looks great! Do you think we should clean up our Google Drive? I had created a bunch of empty files before to make sure we had at least 100 so it would show 100 in the printed `n` at the top of the `tibble` our README, but we don't really need that any more since it just lists everything on the drive.

jennybc:
Thanks!

I see no reason to clean up our Drive right now. I kind of like that there are more than 100 files there, as it provides evidence that we are traversing pages.

Sidebar: BTW, I find that requesting page size larger than 100 doesn't cause an error but also has no effect. It just seems to get set to the maximum of (100, your page size).

jennybc:
About to merge. I learned more about the page size thing: #111

Issue #49, opened 2017-05-23
Download a file


mramos-stellar:
How do you download the file with the use of the already authenticated client? It's not indicated in the functionalities. Is this supported already?

LucyMcGowan:
This isn't formally supported yet (but will be at some point). In the meantime, if you need to do this now I can sketch out some code that could get you started. What type of file are you attempting to download?

mattwilliamson13:
I am also hoping to download a file (primarily spatial files like rasters and shapefiles) into R memory and then run subsequent analyses on it. I am able to see the file via drive_list, but cannot seem to figure out how to download it.

mramos-stellar:
I was supposed to try and add a new function in your code but didn't have time to learn through the interconnected packages and files. :) However, here's a function I wrote that uses your package.

Thanks for this package by the way.



```
#' Download files from Google Drive Folder
#'
#' uses tidyverse/googledrive library for the main listing and authentication handled
#' uses httr library for getting the file download since it's not yet supported by tidyverse/googledrive
#'
#' \code{download_files_from_folder} downloads files from Google Drive folder
#' @author Mark Ramos <mark.ramos@stellarloyalty.com>
#'
#' @param folder_name the name of the Google Drive folder that contains the files you want to download
#' @param query this can be used to customize the query you want; see searching for files guide in Google Drive
#' @param destdir the destination directory on where to store the files to download
#'
#' @export
#' @import googledrive
#' @import httr
#' @import readr
#' @import xml2
#'

download_files_from_folder <- function(folder_name = NULL, query = NULL, destdir = NULL){
  folder <- drive_list(q = paste("name='",folder_name, "'" ,sep = ""))
  if(!is.null(folder)){
    if(is.null(query)){
      query = paste("'",folder$id,"' in parents and trashed=false", sep = "")  
    }
    dumps <- drive_list(q = query)
    for(i in 1:nrow(dumps)){
      auth <- drive_auth()
      dump <- dumps[i,]
      f <- drive_file(dump$id)
      name <- dump$name
      
      if(!is.null(destdir)){
        if(!dir.exists(destdir)){
          dir.create(destdir)
        }
        destination = paste(destdir,name, sep = "/")
      }
      if(!file.exists(destination)){
        print(paste("Downloading File:",name, "with id", f$id))
        
[...]

PR #112, opened 2017-06-26
drive download (#49)

LucyMcGowan:
@jennybc I've recreated the previous pull after adding in the changes from master - let me know what you think!


R/drive_download.R
```diff
@@ -0,0 +1,72 @@
+#' Download a file from Google Drive.
+#'
+#' @template file
+#' @param out_path Character. Path for output file. Google files can
+#'   be downloaded to types specified [here](https://developers.google.com/drive/v3/web/manage-downloads).
```
jennybc: I'm weird about links but I try to avoid vague link text like "here". I try to make it clear where I'm sending people. So either use angle brackets and reveal the URL or come up with descriptive text.

BTW please call me out on same if you see it elsewhere.

R/drive_download.R
```diff
@@ -0,0 +1,72 @@
...
+#' @param overwrite A logical scalar, do you want to overwrite a file that already
+#'    exists locally?
+#' @template verbose
+#'
+#' @examples
+#' \dontrun{
+#' ## Save "chickwts.csv" to the working directory as "chickwts.csv".
+#' drive_download(file = "chickwts.csv", out_path = "chickwts.csv")
+#'
+#' ## Save Google Sheet named "chickwts" to the working directory as
```
jennybc: Maybe say "export" so it's more clear how this differs from the scenario just above?

It might be nice (and good practice/check for us) to add a Google Doc / Word example.

R/drive_download.R
```diff
@@ -0,0 +1,72 @@
...
+#' drive_download(file = "chickwts", out_path = "chickwts.csv")
+#'}
+#' @export
+drive_download <- function(file = NULL,
+                           out_path = NULL,
+                           overwrite = FALSE,
+                           verbose = TRUE) {
+  in_file <- as_dribble(file)
+
+  if (nrow(in_file) == 0L) {
```
jennybc: We do have `confirm_single_file()` and `confirm_some_files()` for this sort of check. I propose we use everywhere possible so that our messaging becomes more uniform.

jennybc: Oh I see you do  just below. I think it's OK to let things error down there if nothing is found.

LucyMcGowan: Oh yes I forgot about `confirm_some_files`! Is it weird to use `confirm_some_files()` and then `confirm_single_file()`? I previously just had `confirm_single_file()` and then felt like the message wasn't very informative if the file actually didn't exist.

R/drive_download.R
```diff
@@ -0,0 +1,72 @@
...
+                                            fileId = in_file$id))
+  } else {
+    ## TODO check that extension is actually allowed per
+    ## https://developers.google.com/drive/v3/web/manage-downloads
+    mime_type <- drive_mime_type(ext)
+    request <- generate_request(endpoint = "drive.files.export",
+                                params = list(fileId = in_file$id,
+                                              mimeType = mime_type ))
+  }
+
```
jennybc: What if you switched these around, so you don't have to negate the condition inside `if()`?

R/drive_download.R
```diff
@@ -0,0 +1,72 @@
...
+  in_file <- confirm_single_file(in_file)
+
+  if (is.null(out_path)) {
+    stop("You must specify `out_path`.")
+  }
+
+  mime_type <- in_file$files_resource[[1]]$mimeType
+
+  ## it seems that it needs an extenstion to save properly
```
jennybc: This is mildly surprising to me, but you've really convinced yourself it's a hard requirement?

R/drive_download.R
jennybc: Perhaps that means we should improve the messaging in those checkers. Or allow a message to be provided in the call, so that you can override with something more informative, when possible and important? You could open an issue for that, but go ahead and simplify things here.

R/drive_download.R
LucyMcGowan: If it doesn't have an extension it seems to just save as a .txt file, but maybe that's fine? I need to have a `mimeType` in order to export the Google types, so I was using this to determine that, but I can do that in a different way if you think that'd be better? Or only require it if the file is a google type?

LucyMcGowan:
@jennybc I've done something strange that is making my new commits not show up here 🙈 (this shows 3 commits, but I have 7 if you actually click on the branch) would you be alright with me closing this and reopening a new one on the same branch?

R/drive_download.R
jennybc: Re: files that start as Google native format. I think you should have a default export format for each. I think we've discussed that elsewhere, in fact, in case you remember where? Given no extra info, fall back to the default. Probably Word, Excel, PowerPoint.

You can definitely create files on your system with no extension. Just do `touch foo` in a shell. So where is the required / forced extension coming from?

jennybc:
It feels like this also needs a `type` argument. Because it feels like both of these should work and they should do the same thing.

```r
drive_download(file = "chickwts", type = "xlsx")
drive_download(file = "chichwts", out_path = "chickwts.xlsx")
```

cc @LucyMcGowan

PR #114, opened 2017-06-26
Add drive_download; closes #49

### LINKED ISSUE #49: Download a file

LucyMcGowan:
* this is a redo of #112


LucyMcGowan:
@jennybc this has a few superfluous commits because I was trying to figure out what was going on with the webhooks earlier, but I'll delete those when we squash :) - I think this is ready for another 👀 I added in the `type` parameter and updated the examples. I've also changed how the extension works, adding the extension specified in `type` if it wasn't specified in the Google Drive name.

R/drive_download.R
```diff
@@ -2,8 +2,10 @@
 #'
 #' @template file
 #' @param out_path Character. Path for output file. Will default to its Google Drive
-#'   name. Google files must include the intended extension and can be downloaded to types
+#'   name. Google files must include the intended extension. This can be specified by the
+#'   `out_path` or the `type` parameter. Google files can be downloaded to types
 #'   specified in the [Drive API documentation](https://developers.google.com/drive/v3/web/manage-downloads).
+#' @param type Character. Extension you would like for the local file.
```
jennybc: Could this be another case of "anything that `drive_mime_type()` can handle? Yeah, I see it will almost always be a file extension, but I don't see any downside to running this through `drive_mime_type()`. If you agree, I have wording to that effect in some other help file, which you could search for.

The wording re `out_path` still seems a bit confusing. You don't have to include the file extension or even `out_path` at all, if you specify `type`, right? Also I suspect "Google files" is not yet clear enough that it means Sheets, Docs, and the like.

R/drive_download.R
```diff
@@ -17,13 +19,22 @@
...
+#'
 #' ## Export a Google Document named "foobar" to the working directory as
-#' ## "foobar.docx"
+#' ## "foobar.docx".
 #' drive_download(file = "foobar", out_path = "foobar.docx")
+#'
+#' ## This will also export a Google Document named "foobar" to the working
+#' ## directory as "foobar.docx".
+#' drive_download(file = "foobar", type = "docx")
 #'}
```
jennybc: I like the examples.

jennybc:
I think you need to get `foobar.docx` out of this branch?

R/drive_download.R
LucyMcGowan: right now I use it to append to the file name, so it must be an extension, but I could change that. Do you think it'd be worthwhile to create an inverse function for `drive_mime_type()` that takes an accepted mime type or extension and returns the extension?

LucyMcGowan:
This got buried in the outdated comments so sticking this here @jennybc:
> right now I use the `type` parameter to append to the file name, so it must be an extension, but I could change that. Do you think it'd be worthwhile to create an inverse function for drive_mime_type() that takes an accepted mime type or extension and returns the extension?

data-raw/export-mime-type-defaults.csv
```diff
@@ -0,0 +1,7 @@
+mime_type_google, mime_type_local
+application/vnd.google-apps.document,application/vnd.openxmlformats-officedocument.wordprocessingml.document
+application/vnd.google-apps.drawing,image/png
+application/vnd.google-apps.form,application/zip
+application/vnd.google-apps.presentation,application/vnd.openxmlformats-officedocument.presentationml.presentation
+application/vnd.google-apps.script,application/vnd.google-apps.script+json
+application/vnd.google-apps.spreadsheet,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
```
jennybc: This defines a default export MIME type for each native Google file type.

R/drive_download.R
```diff
@@ -43,49 +43,104 @@ drive_download <- function(file = NULL,
   file <- as_dribble(file)
   file <- confirm_single_file(file)
 
+  ## preserve extension from out_path, before possible override by file$name
+  ext <- file_ext_safe(out_path)
```
jennybc: I think we only want to consult the extension if it came directly from `out_path`, not if it's from `file$name`. Why? One example is the upload of `foo.csv` into a Sheet. It will still be named `foo.csv` (a regrettable choice, yes, but possible). But we don't want this unfortunate name to drive the download file type, which should default to xlsx for a Sheet. We should discuss....

R/drive_download.R
```diff
@@ -43,49 +43,104 @@ drive_download <- function(file = NULL,
...
+##  Google Sheet --> MS Excel
+## Google Slides --> MS PowerPoint
+get_export_mime_type <- function(mime_type) {
+  m <- .drive$translate_mime_types$mime_type_google == mime_type &
+    is_true(.drive$translate_mime_types$default)
+  if (!any(m)) {
+    stop(glue("Not a recognized Google MIME type:\n{mime_type}"), call. = FALSE)
+  }
+  .drive$translate_mime_types$mime_type_local[m]
+}
```
jennybc: This is where I use the default export MIME types.

R/drive_download.R
```diff
@@ -43,49 +43,104 @@ drive_download <- function(file = NULL,
...
+  m <- mime_tbl$mime_type == mime_type
+  if (sum(m) != 1) return(out_path)
+
+  ext <- mime_tbl$ext[m]
+  ext_orig <- file_ext_safe(out_path)
+  if (!identical(ext, ext_orig)) {
+    out_path <- paste(out_path, ext, sep = ".")
+  }
+  out_path
+}
```
jennybc: This is a slightly ugly helper. If we keep it, perhaps you should write that function to get extension from MIME type and then use it here. So break this up into two functions: one that gets extension, one that applies it. Move the first into the MIME type file (or path-utils.R?) and the second into path-utils.R. Regardless, this needs some tests.

jennybc:
This does some of the stuff I've discussed. See what you think. We can discuss on tomorrow's call.

R/drive_download.R
jennybc: We need to deal with the fact that the tibble returned by `drive_mime_type()` can have multiple extensions for a MIME type.

R/drive_download.R
jennybc: If we keep this, we need to make it easier for people to see what file types are available for each Google type and what our chosen default is.

LucyMcGowan:
@jennybc I think this is ready for a final 👀

PR #117, opened 2017-06-28
Update Mime type handlers

LucyMcGowan:


R/drive_mime_type.R
```diff
@@ -10,7 +10,7 @@
 #'   * File extensions, such as "pdf", "csv", etc.
 #'   * MIME types accepted by Google Drive (these are simply passed through).
 #'
-#' @description If no input is provided, function returns the full table used
+#' @description If `type = all`, function returns the full table used
```
jennybc: I don't think we can do it this way. Because `"all"` is character input that, in theory, could be legitimate input. Now, I grant you that `"all"` is not a file extension nor a human-friendly nickname nor an actual MIME type. And yet I still don't think we should do this. I will think on it.

tests/testthat/test-drive_mime_type.R
```diff
@@ -1,9 +1,21 @@
...
+test_that("drive_mime_type() returns NULL if no input", {
+  expect_null(drive_mime_type())
+})
+
+test_that("drive_mime_type(expose()) returns the full tibble", {
+  expect_identical(
+    drive_mime_type(expose()),
+    .drive$mime_tbl
+  )
+>>>>>>> a32282f1dc988eb917681f52d62452e11b046985
```
jennybc: oops

LucyMcGowan: yikes 😬

jennybc:
Will you merge master in and then I'll review?

LucyMcGowan:
👌 @jennybc I think this is ready for a 👀.

tests/testthat/test-drive_mime_type.R
```diff
@@ -11,11 +11,6 @@ test_that("drive_mime_type(expose()) returns the full tibble", {
   )
 })
 
-test_that("drive_mime_type() returns NULL if input NULL", {
-  out <- drive_mime_type()
-  expect_null(out)
-})
-
```
jennybc: This already exists just above.

PR #118, opened 2017-06-28
reconcile consistency of function signatures

LucyMcGowan:
I've updated all parameters that reference a `folder` or `path` to be called `path`. This includes:
* `drive_download()` (what used to be `out_path` -> `path`)
* `drive_mkdir()`
* `drive_upload()` (what used to be `folder` -> `path`)
* `drive_mv()` (what used to be `folder` -> `path`)
* `drive_ls()`


R/drive_auth.R
```diff
@@ -134,13 +134,25 @@ drive_auth <- function(token = NULL,
 
     drive_token <- try(suppressWarnings(readRDS(token)), silent = TRUE)
     if (inherits(drive_token, "try-error")) {
-      stop(glue("Cannot read token from alleged .rds file:\n{token}"))
+      stop(
+        glue("Cannot read token from alleged .rds file:\n{token}"),
+        call. = FALSE
```
jennybc: There's basically a tidyverse policy to use `call. = FALSE`, FYI.

man-roxygen/path.R
```diff
@@ -1,4 +1,4 @@
-#' @param path Drive File, something that identifies the where the file should
-#'   live on your Google Drive. Can be an actual path (character), a file id marked with [as_id()], or
-#'   a [`dribble`].
+#' @param path Drive File, something that identifies where the file should live
+#'   on your Google Drive. Can be an actual path (character), a file id marked
+#'   with \code{\link{as_id}()}, or a \code{\link{dribble}}.
```
jennybc: I think we learned the hard way that templates can't use the Markdown markup.

PR #121, opened 2017-06-29
Add trash functions (closes #104)

### LINKED ISSUE #104: Delete vs put in trash
I wonder if we should offer `drive_trash()` instead of `drive_delete()`.

And then have a function to empty the trash.

And `drive_untrash()` to pull something back out of the trash.

The browser UI makes it much harder to truly delete things than we currently do. I don't want to make it easy for people to shoot themselves in the foot.

LucyMcGowan:


R/drive_delete.R
```diff
@@ -1,5 +1,6 @@
 #' Delete file from Google Drive.
 #'
+#' @description Caution: this will permanently delete your files.
```
jennybc: I think we should only use the `@description` tag when necessary, such as when we're trying to include text after some bullet points in the same section.

R/drive_delete.R
```diff
@@ -32,6 +33,5 @@ delete_one <- function(id) {
     endpoint = "drive.files.delete",
     params = list(fileId = id)
   )
-  response <- make_request(request)
-  process_response(response)
+  do_request(request)
```
jennybc: I propose that we NOT indulge in `do_request()` but rather keep doing `make_request()` then `process_response()`. On principle, I think future debugging will be easier if we keep individual steps.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
+#' Trash file on Google Drive.
```
jennybc: I just did a massive removal of "Google" from "Google Drive" in help files, because it comes up so much and I think it's basically implied at this point. We will spell it out in some key places, but otherwise let's just say "Drive".

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+#'
+#' @template file
+#' @template verbose
+#'
+#' @return Logical, indicating whether the trashing succeeded.
+#' @export
+#' @examples
+#' \dontrun{
+#' ## Trash a single file in the trash.
```
jennybc: "Put a single file in the trash"?

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
+#' Trash file on Google Drive.
+#'
+#' @template file
+#' @template verbose
+#'
+#' @return Logical, indicating whether the trashing succeeded.
+#' @export
+#' @examples
+#' \dontrun{
```
jennybc: Propose you add an upload command so we can actually run examples without writing extra code. See the help for `drive_cp()` or `drive_upload()` for inspiration.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+  if (verbose) {
+    if (any(out)) {
+      successes <- glue_data(trash_file[out, ], "  * {name}: {id}")
+      message(collapse(c("Files trashed:", successes), sep = "\n"))
+    }
+    if (any(!out)) {
+      failures <- glue_data(trash_file[!out, ], "  * {name}: {id}")
+      message(collapse(c("Files NOT trashed:", failures), sep = "\n"))
+    }
+  }
```
jennybc: Do you think it's possible to fail and get this far when we've failed to trash the file? I guess I could believe it if we were using `purrr::safely()` to wrap `trash_one` and, perhaps, we should be. I suppose you could fail it you don't own the file. Maybe you should explore that scenario to think about how these vectorized functions should work when failure is possible and doesn't raise an error.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+  invisible(out)
+}
+
+trash_one <- function(id, trash = TRUE) {
+  request <- generate_request(
+    endpoint = "drive.files.update",
+    params = list(fileId = id,
+                  trashed = trash)
+  )
+  proc_res <- do_request(request, encode = "json")
```
jennybc: Ditto the `do_request()` thing said elsewhere.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+  identical(proc_res$id, id)
+}
+
+#' Untrash file on Google Drive.
```
jennybc: These could conceivably be documented together. That would make the examples nice.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+#' @examples
+#' \dontrun{
+#' ## Untrash a single file in the trash.
+#' drive_untrash("chickwts.csv")
+#'
+#' ## Untrash multiple files in trash.
+#' drive_untrash(c("abc", "def"))
+#' }
+drive_untrash <- function(file = NULL, verbose = TRUE) {
+  if (is.character(file)) {
```
jennybc: Needs to use the new `is_path()` helper. As is, doesn't anticipate a drive id.

jennybc: I think `file` should really be a `file` they way it is elsewhere, i.e. user should call `drive_search()` if that's how they want to find the file. We should just make sure it's in the trash or return early.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+#' @template dribble-return
+#' @export
+drive_view_trash <- function() {
+  drive_search(q = "trashed = true")
+}
+
+#' Empty Google Drive Trash.
+#'
+#' @template verbose
+#' @description Caution: this will permanently delete files in your
```
jennybc: Should go above the doc for `verbose` and does not need explicit `@description` tag.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+#' @export
+drive_empty_trash <- function(verbose = TRUE) {
+  files <- drive_view_trash()
+  del <- drive_delete(files, verbose = FALSE)
+  n <- sum(del)
+  if (verbose) {
+    if (n > 0L) {
+    message(glue("You have successfully deleted {n} files from your Google Drive trash."))
+    } else {
+      message("Your trash was already empty.")
```
jennybc: I don't quite follow the logic. This would seem linked to the length of `files`.

R/drive_trash.R
```diff
@@ -0,0 +1,111 @@
...
+  }
+  invisible(out)
+}
+
+#' View files in Google Drive Trash.
+#' @template dribble-return
+#' @export
+drive_view_trash <- function() {
+  drive_search(q = "trashed = true")
+}
```
jennybc: I wonder if we should do this with `drive_trash()` (called with no arguments) or `drive_trash(expose())` (called with our "show me this thing" sentinel? And not have a single purpose function.

jennybc: OTOH it undermines the predictability of `drive_trash()`.

tests/testthat/test-drive_trash.R
```diff
@@ -0,0 +1,52 @@
...
+
+## NOTE if you do not currently have the files needed,
+## change run & clean below to TRUE to create files needed
+## (CAUTION, this will delete files that will interfere)
+
+
+nm_ <- nm_fun("-TEST-drive-trash")
+
+run <- FALSE
+clean <- FALSE
```
jennybc: I'm just starting to have two blocks like so:

```
## clean
if (FALSE) {
  # delete stuff
}

## setup
if (FALSE) {
  # create stuff
}
```

because in reality I just interactively execute those lines when needed. The nesting and objects `run` and `clean` seem like something we can abandon.

tests/testthat/test-drive_trash.R
```diff
@@ -0,0 +1,52 @@
...
+  }
+  drive_mkdir(nm_("foo"))
+}
+
+test_that("drive_trash() moves object to the trash", {
+
+  skip_on_travis()
+  skip_on_appveyor()
+
+  expect_true(drive_trash(nm_("foo")))
```
jennybc: How is `foo` in the trash at this point?

tests/testthat/test-drive_trash.R
```diff
@@ -0,0 +1,52 @@
...
+  if (clean) {
+    del <- drive_delete(nm_("foo"), verbose = FALSE)
+  }
+  drive_mkdir(nm_("foo"))
+}
+
+test_that("drive_trash() moves object to the trash", {
+
+  skip_on_travis()
+  skip_on_appveyor()
```
jennybc: You probably want an `on.exit()` here that undoes anything you do. So, untrashes the test file.

jennybc: Maybe this and test below should be merged, so they can work together, actually.

R/drive_trash.R
LucyMcGowan: I think these would be the same? Both the length of `files` and `sum(del)` - would it be better to have `n` be the length rather than the sum?

R/drive_trash.R
LucyMcGowan: In retrospect I can't think of any ways to fail...should I remove that then?

LucyMcGowan:
@jennybc I think this is ready for another 👀.

R/drive_trash.R
jennybc: I think, unless we've taken measure to catch errors, we will have errored by now if the deletion didn't succeed. So, yeah, I would remove the complexity here.

R/drive_trash.R
jennybc: I think the function is easier to follow if you message and return for the "no files in trash" scenario right away. I think  the current version does the right thing, but you have to work it all out in your head to be sure (what if the trash starts out empty? what if it's not empty but we somehow didn't delete it all? what if...?).

R/drive_rm.R
```diff
@@ -32,6 +33,5 @@ delete_one <- function(id) {
     endpoint = "drive.files.delete",
     params = list(fileId = id)
   )
-  response <- make_request(request)
-  process_response(response)
+  do_request(request)
```
jennybc: Can we standardize on `make_request()` + `process_response()`? I know it seems silly here, but I think it's better to be consistent. One of us will read this one day and wonder if there's a reason it's different.

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
+#' Trash file on Drive.
```
jennybc: Move Drive files to the trash

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+#' @template file
+#' @template verbose
+#'
+#' @return Logical, indicating whether the trashing/untrashing succeeded.
```
jennybc: vector?

R/drive_rm.R
LucyMcGowan: oh I forgot that I had changed this before! I'll put it back to `make_request()` and `process_response()`.

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+#'
+#' ## Put multiple files in trash.
+#' file_a <- drive_upload(system.file("DESCRIPTION"), "file_a")
+#' file_b <- drive_upload(system.file("DESCRIPTION"), "file_b")
+#' drive_trash(c("file_a", "file_b"))
+#' drive_untrash(c("file_a", "file_b"))
+#' }
+drive_trash <- function(file = NULL, verbose = TRUE) {
+  trash_file <- as_dribble(file)
+  if (!some_files(trash_file) && verbose) {
```
jennybc: `no_file()` is nicer than `!some_files()`

Should this return here? And invisibly return `logical(0)`?

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+  if (verbose) {
+    if (any(out)) {
+      successes <- glue_data(trash_file[out, ], "  * {name}: {id}")
+      message(collapse(c("Files trashed:", successes), sep = "\n"))
+    }
+  }
+  invisible(out)
+}
+
+trash_one <- function(id, trash = TRUE) {
```
jennybc: Can we call this `set_trash_one()` or `toggle_trash_one()`? I had to read the code to figure out that it does not just put things in the trash.

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+  identical(proc_res$id, id)
+}
+
+#' @rdname drive_trash
+#' @export
+drive_untrash <- function(file = NULL, verbose = TRUE) {
+  if (is_path(file)) {
+    paths <- collapse(file, sep = "|")
+    file <- drive_find(paths, q = "trashed = true")
+  }
```
jennybc: Is this going to work if one of the paths is `path/to/a/file`?

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+  trash_file <- as_dribble(file)
+  if (!some_files(trash_file) && verbose) {
```
jennybc: `no_file()`

R/drive_trash.R
LucyMcGowan: 😬 no I don't think so - I would have used `drive_paths()` but right now we can't pass `q` and we default to not looking in the trash.

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+    message(glue("No such files found to trash"))
+    return(invisible(FALSE))
```
jennybc: I *think* this should be `logical(0)`? The principle would be that length of output = length of input.

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  out <- purrr::map_lgl(trash_file$id, trash_one, trash = FALSE)
+
+  if (verbose) {
+    if (any(out)) {
```
jennybc: This should be symmetric with `drive_trash()` messaging, no?

LucyMcGowan: oops!

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+#' View files in Drive Trash.
+#' @template dribble-return
+#' @export
+drive_view_trash <- function() {
+  drive_find(q = "trashed = true")
+}
+
+#' Empty Drive Trash.
+#' Caution: this will permanently delete files in your
+#'   Drive trash.
```
jennybc: weird line break

R/drive_trash.R
```diff
@@ -0,0 +1,100 @@
...
+  files <- drive_view_trash()
+  n <- nrow(files)
+  if (n == 0L) {
+    message("Your trash was already empty.")
+    return(invisible(NULL))
+  }
+  del <- drive_rm(files, verbose = FALSE)
+  if (verbose) {
+    message(glue("You have successfully deleted {n} file(s) from your Google Drive trash."))
+  }
```
jennybc: Be intentional about what this returns.

LucyMcGowan: what do you think it should return? 🤔 logical like the others?

tests/testthat/test-drive_trash.R
```diff
@@ -0,0 +1,49 @@
...
+
+
+test_that("drive_trash() moves object to the trash and drive_untrash() undoes", {
+
+  skip_on_travis()
+  skip_on_appveyor()
+  skip_if_offline()
+
+  expect_true(drive_trash(nm_("foo")))
+  foo <- drive_find(nm_("foo"), q = "trashed = true")
```
jennybc: This and the `drive_find()` below should be `drive_path()` (at this moment) and `drive_get()` (very soon). Because `drive_find()` will retrieve *all of user's files* in order to fish out this single file. Better to target it.

tests/testthat/test-drive_trash.R
```diff
@@ -0,0 +1,49 @@
+context("Trash files")
+
+## WARNING: this will empty your drive trash. If you do
```
jennybc: Move this down near the actual test and reword to reflect current situation. It's a bit alarming right now, since the test is actually commented out.

R/drive_trash.R
jennybc: OK can you make this as close as possible to what it should be and leave TO DOs and/or open or comment on issues then?

R/drive_trash.R
jennybc: I guess `invisible(TRUE)`. I believe that should also be return value in the empty trash scenario. Basically, unless there's an error, this always returns `invisible(TRUE)`.

R/drive_trash.R
jennybc: Are you sure about the `drive_find()` behaviour? From our code, it's clear our intent is to NOT mess with trash if user's `q` says anything about trash.

LucyMcGowan: It's true that you can pass q to `drive_find()`, but I think the proper way to do this so it could handle a path would be to pass it to `drive_paths()` which can't have a `q` passed at the moment.

PR #122, opened 2017-06-29
beef up tests

LucyMcGowan:
I ran `package_coverage()` locally - we have a little over 80% covered - I noticed a few places where we weren't testing things so I've added those here. I also began migrating away from creating the `chickwts.csv` and using the internal files discussed on Slack to upload 🎉


R/drive_share.R
```diff
@@ -32,7 +32,7 @@ drive_share <- function(file = NULL, role = NULL, type = NULL, ..., verbose = TR
   if (!(role %in% ok_roles)) {
     stop(
       glue(
-        "`role` must be one of the following: {paste(ok_roles, collapse = ", ")}."
+        "`role` must be one of the following: \n{paste(ok_roles, collapse = ', ')}."
       ),
```
jennybc: Not sure if you want to do this here or in another PR.
But note that `glue()` will paste together individual strings, so you can break things up, if needed.
I find right after a `\n` is a nice place, because then the code simulates the way the message will print.
Above, you don't want the space between `:` and `\n`.
Also, let's switch to glue's `collapse()` instead of `paste(..., collapse = ...)`.

PR #141, opened 2017-07-13
split upload into drive_upload() and drive_update() (closes #71)

### LINKED ISSUE #71: drive_upload(): when name or id already exists
* In order to upload with a new mimeType, it seems you have to both upload & change the metadata simultaneously
* It seems the only way to do so is with a multipart upload 
    * don't know how to do that without writing a temporary file
* This was asked about on [SO](https://stackoverflow.com/questions/31080363/how-to-post-multipart-related-content-with-httr-for-google-drive-api).
* Addressed a bit here: https://github.com/hadley/httr/issues/253

LucyMcGowan:


jennybc:
Something's still not quite right here. This may have nothing to do with the code you're touching here, but this is a good place to work it out. Have you successfully uploaded new media into existing file id?

``` r
devtools::load_all(".")
#> Loading googledrive
#> Auto-refreshing stale OAuth token.
target <- drive_upload(R.home('doc/BioC_mirrors.csv'), "jenny-fiddle")
#> File uploaded:
#>   * jenny-fiddle
#> with MIME type:
#>   * text/csv
target <- drive_upload(system.file("DESCRIPTION"), target)
#> Error: Requested parent folder does not exist:
#> jenny-fiddle
```

jennybc:
Think through scenarios where `path` pre-exists vs does not and is a folder vs is a file.

LucyMcGowan:
I have:
```
x <- drive_upload(system.file("DESCRIPTION"), name = "testing-upload-lucy")
#> File uploaded:
#>  * testing-upload-lucy
#> with MIME type:
#>   * text/plain
> x
#> # A tibble: 1 x 3
#>                  name                           id files_resource
#> *               <chr>                        <chr>         <list>
#> 1 testing-upload-lucy 0B0Gh-SuuA2nTWUZPMGI1a3Z3R2s    <list [36]>
> y <- drive_upload(system.file("DESCRIPTION"), name = "testing-upload-lucy", overwrite = TRUE)
#> File uploaded:
#>   * testing-upload-lucy
#> with MIME type:
#>   * text/plain
> y
#> # A tibble: 1 x 3
#>                  name                           id files_resource
#> *               <chr>                        <chr>         <list>
#> 1 testing-upload-lucy 0B0Gh-SuuA2nTWUZPMGI1a3Z3R2s    <list [37]>
```
I think the problem in the one above is that `target` isn't a folder, but when you stick something in the second parameter slot, it expects a folder.

LucyMcGowan:
but maybe I'm doing it wrong the way I have it? It currently searches for the same in a given folder than than for the specific file id, but perhaps the latter makes more sense?

jennybc:
So if `path` already exists but it's a folder, the upload should go through.

But if `path` pre-exists and it's a file and you haven't specified `overwrite = TRUE`, the message should be more informative.

jennybc:
It feels like my example above should work like this:

```r
target <- drive_upload(R.home('doc/BioC_mirrors.csv'), "jenny-fiddle")
#> File uploaded:
#>   * jenny-fiddle
#> with MIME type:
#>   * text/csv
target <- drive_upload(system.file("DESCRIPTION"), target)
#> NOPE, TARGET FILE ALREADY EXISTS AND YOU HAVEN'T SAID TO OVERWRITE
target <- drive_upload(system.file("DESCRIPTION"), target, overwrite = TRUE)
#> YES THIS SHOULD UPDATE MEDIA FOR TARGET
```

Agree? But I still think there are a lot of scenarios to enumerate and try. I suspect we're still not done.

LucyMcGowan:
So then maybe we would never error that the parent folder does not exist? Right now we check if `path` is a folder & if not error that the path you were hoping to stick something in doesn't exist, but if we change to the suggestion 👆, we would:
(1) check if you gave us a path, if not stick the file in the `root_folder`, if yes go to (2)
(2) check if the path you gave us is a folder, if yes, get the folder id to set as the parent, if no go to (3)
(3) check if the path is a file, if yes, check if `overwrite == TRUE`, if yes, overwrite this id, if no give error "TARGET FILE ALREADY EXISTS AND YOU HAVEN'T SAID TO OVERWRITE 😜"

jennybc:
> So then maybe we would never error that the parent folder does not exist?

But that can't quite be right?

```r
drive_upload(R.home('doc/BioC_mirrors.csv'), "jenny-fiddle", name = "something")
```

should error if there's no pre-existing folder named "jenny-fiddle".

LucyMcGowan:
ah 👌 
(1) check if you gave us a path, if not stick the file in the root_folder, if yes go to (2)
(2) check if the path you gave us is a folder, if yes, get the folder id to set as the parent, if no check if you gave a name, if yes error that the thing in `path` is not a folder if no go to (3)
(3) check if the path is a file, if yes, check if overwrite == TRUE, if yes, overwrite this id, if no give error "TARGET FILE ALREADY EXISTS AND YOU HAVEN'T SAID TO OVERWRITE 😜"

jennybc:
I think we're getting there. But I would think of it more in terms of what user has told you about the target destination (is it a file or a folder? does it pre-exist?) and then work backwards to the procedural stuff vs starting with lots of if/else if/else if/else. Definitely test it all out in a variety of situations and see how it feels. We might not want to literally have all of that in tests or examples, so perhaps park your explorations in a draft article.

LucyMcGowan:
👌 I've played a bit and added a small [article](https://github.com/tidyverse/googledrive/blob/drive_upload/internal/drive_upload.md) to demonstrate where I've currently landed. Not certain about everything (for example if you are trying to overwrite AND change the name, the error is currently about folders, which may be confusing). Let me know what you think.

jennybc:
The article helps immensely! Will read it now.

R/drive_upload.R
```diff
@@ -66,35 +66,49 @@ drive_upload <- function(file = NULL,
...
+      stop(
+        glue(
+          "Requested parent folder does not exist:\n{path$name}"
+        ),
+        call. = FALSE
+      )
+    }
+    if (!overwrite) {
+      stop(
+        glue("File already exists:\n{path$name}"),
```
jennybc: Let's suggest the fix and bulletize the file as we are starting to do everywhere. In fact, that would be good for this entire PR.

> File exists. Use `overwrite = TRUE` to upload new content into this file id.

internal/drive_upload.Rmd
```diff
@@ -0,0 +1,129 @@
+---
+title: "Playing with `drive_upload()`"
+author: "Lucy D’Agostino McGowan"
+date: "7/14/2017"
+output: github_document
+---
+
+```{r}
+library("googledrive")
+library("testthat")
```
jennybc: Not sure of your motivation but you can also set `error = TRUE` globally or chunk wise to render in the presence of errors and reveal them.

R/drive_upload.R
```diff
@@ -66,35 +66,49 @@ drive_upload <- function(file = NULL,
...
+        call. = FALSE
+      )
+    }
+    existing <- path
+    path <- unlist(promote(path, "parents")$parents)
```
jennybc: I wouldn't use `promote()` here. I would get parents "by hand".

jennybc: You also need to worry about parents having length greater than 1. But I actually think we shouldn't be monkey with the parents if an existing non-folder file target anyway. I'm about to make some higher level comments.

R/drive_upload.R
```diff
@@ -66,35 +66,49 @@ drive_upload <- function(file = NULL,
   path <- path %||% root_folder()
   path <- as_dribble(path)
   confirm_single_file(path)
+  existing <- NULL
```
jennybc: It doesn't feel correct that we've got `existing` and `up_id` floating around. I suspect something can still be simplified.

R/drive_upload.R
```diff
@@ -66,35 +66,49 @@ drive_upload <- function(file = NULL,
...
+      existing <- drive_find(q = qq)
 
-  if (nrow(existing) > 0) {
-    out_path <- unsplit_path(path$name %||% "", name)
-    if (!overwrite) {
-      stop(glue("Path already exists:\n{out_path}", call. = FALSE))
-    }
-    if (nrow(existing) > 1) {
-      stop(glue("Path to overwrite is not unique:\n{out_path}", call. = FALSE))
+      if (nrow(existing) > 0) {
```
jennybc: If this survives, you could use the `some_files()` helper.

jennybc:
This is a high-level comment, because I'm still trying to sort out my PR.

In this neighbourhood:

```r
if (is_path(path)) {
    if (is.null(name) && drive_path_exists(append_slash(path))) {
      path <- append_slash(path)
    }
    path_parts <- partition_path(path, maybe_name = is.null(name))
    path <- path_parts$parent
    name <- name %||% path_parts$name
    ## if we've provided name of existing file via path
    ## is path now null and name holds the name?
  }
```

It feels like my comment should be true (or should it?). If we could really count on `path` holding parent folder and `name` holding a name by this point, everything below sure would be simpler. The sooner and simper you can figure out that we're (potentially) updating an existing file id, the better. FWIW I also don't think you should be able to rename it at the same time. Force people to use `drive_rename()` for that.

R/drive_upload.R
jennybc: To be very clear, if we're updating media for existing id, we should NOT be touching parents. It would be easy to do more harm than good. In which case, I don't see why we need to store the parents.

R/drive_upload.R
jennybc: In fact, this whole little bit could be simplified. Why are we checking if rows > 0 then if rows > 1? I realize it may have been me who wrote this!

jennybc:
Also, if we are updating existing file id, the message should say "File updated with new media" or something like that, instead of the same "File uploaded" you see for a new file id.

jennybc:
And maybe include the file id, the way we do in lots of other message elsewhere?

internal/drive_upload.Rmd
```diff
@@ -0,0 +1,129 @@
...
+```
+
+If I try to give it a new name, it should error.
+```{r}
+expect_error(
+  c <- drive_upload(system.file("DESCRIPTION"), b, name = "new_name", overwrite = TRUE),
+  "Requested parent folder does not exist"
+)
+```
+*hmm is this a sensible error? it definitely makes sense in some circumstances but...*
```
jennybc: Yeah, I think it should never get to this point. If `b` is a pre-existing non-folder file, we're going to ignore `name`.

LucyMcGowan: I think I agree - then we are back to never really erroring if we don't have a folder (essentially, the circumstance would be that `confirm_single_file(path)` would not be able to confirm that, so the error would state:
```r
#> Input does not hold exactly one Drive file:
#> path
```

jennybc: > I think I agree - then we are back to never really erroring if we don't have a folder (essentially, the circumstance would be that confirm_single_file(path) would not be able to confirm that).

I don't understand.

LucyMcGowan: If we allow you to pass a non-folder to the `path` parameter, and if you *also* give a `name` we default to ignoring the name, then (as it is currently set up) I don't think it'd be possible to hit an error that what you have provided isn't a folder, since the only other way to provide a non-folder would be to provide something that doesn't exist at all on your Drive, in which case it would be caught by the `confirm_single_file()`. The `confirm_single_file()` checker doesn't have an error pertaining to folders specifically, it just states the thing you gave it isn't a Drive file (maybe this is okay?).

LucyMcGowan: also, in this framework, if `b` is a pre-existing non-folder file & we have `name` it'd be simple to allow it (and allow it to actually change the name) rather than ignoring it & requiring the user to use `drive_rename()` (but perhaps this is confusing?)

jennybc: I can't hold all the scenarios in my head right now, but trust you are working through them.

I'll just restate one of my original assertions, i.e. that this should tell you the folder "jenny-fiddle" doesn't exist:

```
drive_upload(R.home('doc/BioC_mirrors.csv'), "jenny-fiddle", name = "something")
```

I worried because "never really erroring if we don't have a folder " suggests somehow we wouldn't catch such a thing. But I'm probably just not following it all.

I continue to think that `drive_upload()` should NOT support "new media AND new name" at the same time.

LucyMcGowan: ah okay, so then this wouldn't hold:
> Yeah, I think it should never get to this point. If `b` is a pre-existing non-folder file, we're going to ignore `name`.
The way we have the function now, we check if `jenny-fiddle` exists any where on drive (but using the `confirm_single_file()` function:
* if it does NOT exist at all, you will receive the generic `confirm_single_file()` error message
* if it does exist, but as a file, it errors because you provided a `name`, so we are assuming `jenny-fiddle` was intended to be a folder
* if it does exist as a folder you go on your happy way
I'm sorry to pull you into the weeds on this - I just want to make sure the functionality is as you would expect (it feels logical to me at the moment, but probably because I wrote it as such 😬) .

jennybc: So maybe we need `confirm_single_folder()`.

LucyMcGowan: I think maybe the problem is more that I'm not sure how to structure it so that this errors that `jenny-fiddle` isn't a folder:
```
drive_upload(R.home('doc/BioC_mirrors.csv'), "jenny-fiddle", name = "something")
```
but this just ignores name:

```
drive_upload(R.home('doc/BioC_mirrors.csv'), b,  name = "something")
```
since `jenny-fiddle` and `b` are treated the same by the time they reach the erroring portion. Does that make sense?

LucyMcGowan:
👌 I've made a bit of progress here @jennybc & updated an [article](https://github.com/tidyverse/googledrive/blob/drive_upload/internal/drive_upload.md) if you want to take a peek.

R/drive_upload.R
```diff
@@ -8,7 +8,9 @@
 #' @param name Character, name the file should have on Google Drive if not
 #'   specified in `path`. Will default to its local name.
 #' @param overwrite A logical scalar, do you want to overwrite a file already on
-#'   Google Drive, if such exists?
+#'   Google Drive, if such exists? If you intend to overwrite an existing
+#'   `dribble` by specifying it using `path` and `overwrite = TRUE`, leave
+#'   `name` `NULL`. You can rename this file using `drive_rename()`.
```
jennybc: Use the internal-link-friendly syntax: `[drive_rename()]`

LucyMcGowan:
👌 @jennybc I've split into `drive_upload()` and `drive_update()` - let me know what you think.

R/drive_update.R
```diff
@@ -1,19 +1,23 @@
...
+#' Update an existing Drive file
+#'
+#' @seealso Wraps the
+#' [drive.files.update](https://developers.google.com/drive/v3/reference/files/update)
+#' endpoint. In particular, does [media upload](https://developers.google.com/drive/v3/web/manage-uploads).
 #'
 #' @param file Character, path to the local file to upload.
 #' @template path
-#' @param ... Query parameters to pass along to the API query.
+#' @param ... Parameters to pass along to the API query. NOT IMPLEMENTED YET.
```
jennybc: Some of these will go in the body, once we truly support metadata updates (https://github.com/tidyverse/googledrive/issues/148).

R/drive_update.R
```diff
@@ -53,20 +57,10 @@ drive_update <- function(file = NULL,
...
-    }
-  } else {
-    stop("The file doesn't seem to have updated.", call. = FALSE)
+  if (verbose) {
+    message(glue("\nFile updated:\n  * {out$name}: {out$id}"))
   }
-
-  invisible(updated_file)
+  invisible(out)
 }
```
jennybc: I made this look more like its friends, e.g. `drive_copy()`. I think it's another case of "if we failed, we would know by now ... or, at least, this id check doesn't prove anything". Agree?

tests/testthat/test-drive_update.R
```diff
@@ -22,23 +22,35 @@ test_that("drive_update() updates file", {
...
+  tmp <- tempfile()
+  now <- as.character(Sys.time())
+  writeLines(now, tmp)
+
+  out <- drive_update(tmp, updatee)
+  expect_identical(out$id, updatee$id)
+  drive_download(updatee, tmp, overwrite = TRUE)
+  now_out <- readLines(tmp)
+  expect_identical(now, now_out)
 })
```
jennybc: I made this a more genuine test of whether we have, in fact, updated anything at all!

R/drive_update.R
LucyMcGowan: Yes! I believe that is correct 👌

jennybc:
I think the interface of `drive_update()` should change: the Drive file we are updating should be the first argument. And the file to upload should be a later argument, named `media` (??) or something like that. Related to https://github.com/tidyverse/googledrive/issues/148, but you don't need to close that issue to do this.

jennybc:
I think this is ready, other than my comment above about the signature of `drive_update()`.

LucyMcGowan:
Ah I agree, so that `drive_update()` can be piped into. Should we also change `drive_upload()` to have the argument `media` instead of `file`?

jennybc:
> Should we also change `drive_upload()` to have the argument media instead of file?

I have no strong feelings, but yeah that sounds right. It would, for example, probably allow you to share a template re: docs.

LucyMcGowan:
Or do you think something like `local_path` would work? Or is that too long?

jennybc:
I think I favour media because of overlap with API docs.

jennybc:
It might also generalize better in the future, for example if we support upload from R object to Drive file.

PR #144, opened 2017-07-18
Retain dribble class after some dplyr calls

jennybc:
Goal: retain the dribble class after, e.g. row filtering or sorting via dplyr verbs. Strip the dribble class after, e.g. renaming or mutating that leaves an invalid dribble. Have the minimal dependency on dplyr. I wanted no connection to dplyr, but it appears that "Suggests" is absolutely required.

Does this look like the right way to do this @hadley @lionel- ? It differs from the dtplyr example in a few respects.


R/dribble.R
```diff
@@ -85,7 +85,7 @@ maybe_dribble <- function(x) {
   if (all(ok)) {
     new_dribble(x)
   } else {
-    x
+    structure(x, class = class(tibble::tibble()))
```
jennybc: This little exercise revealed that some dplyr verbs don't strip the dribble class, even when they should! Hence this. Is this the best way to get and apply the class attribute for a plain vanilla tibble? Or should I write `as_tibble.dribble()` 🙄?

lionel-: `as_tibble()` may be a bit clearer

R/dribble.R
```diff
@@ -85,7 +85,7 @@ maybe_dribble <- function(x) {
   if (all(ok)) {
```
lionel-: Using `all()` on a scalar logical is a bit misleading for the reader

PR #151, opened 2017-07-20
Update message handling (fixes #149, fixes #32, fixes #113)

### LINKED ISSUE #32: Uniform handling of messages
We should work out some policies about messages, warning, and errors. Re: the form of the message and the mechanics of how we'll make them. There's lots of diversity right now.

### LINKED ISSUE #113: Update messages in confirm functions
* Look into updating `confirm_single_file()` to allow for custom messages as mentioned in #112

> Perhaps that means we should improve the messaging in those checkers. Or allow a message to be provided in the call, so that you can override with something more informative, when possible and important?

### LINKED ISSUE #149: Favor path over name in drive_cp(), drive_upload() & friends
Our docs claim:
```
#' @param name Character, new file name if not specified as part of `path`. Any
#'   name obtained from `path` overrides this argument. Defaults to "Copy of
#'   {CURRENT-FILE-NAME}."
```
but the code seems to imply exactly the opposite 🤔:
``` 
 if (is_path(path)) {
    if (is.null(name) && drive_path_exists(append_slash(path))) {
      path <- append_slash(path)
    }
    path_parts <- partition_path(path, maybe_name = is.null(name))
    path <- path_parts$parent
    name <- name %||% path_parts$name
  }
```

- [x] This can likely be fixed by flipping the last line as so:
```
name <- path_parts$name %||% name
```
*edit: ended up changing the docs instead*
- [x] Add tests checking this behaves as expected

LucyMcGowan:
See

#149 Favor path over name in drive_cp(), drive_upload() & friends

#113 Update messages in confirm functions

#32 Uniform handling of messages


R/drive_mkdir.R
```diff
@@ -5,8 +5,9 @@
 #'   [dribble]. If specified as an actual path, it is best to explicitly
 #'   indicate if it's a folder by including a trailing slash, since it cannot
 #'   always be worked out from the context of the call.
-#' @param name Character. The name of the folder you would like to create, if
-#'   not already specified in `path`.
+#' @param name Character. The name of the folder you would like to create. This
+#'   will force `path` to be treated as a folder, even if it is character and lacks
+#'   a trailing slash.
 #' @template verbose
```
jennybc: Seems identical to something in `drive_cp()`. Opportunity to use a template?

R/drive_mkdir.R
```diff
@@ -54,9 +55,18 @@ drive_mkdir <- function(path = NULL, name = NULL, verbose = TRUE) {
 
   path <- path %||% root_folder()
   path <- as_dribble(path)
-  confirm_single_file(path)
+  if (!single_file(path)) {
```
jennybc: What if file contains zero files? I think we lost a check here.

R/drive_mv.R
```diff
@@ -4,9 +4,9 @@
 #'
 #' @template file
 #' @template path
-#' @param name Character, new file name if not specified as part of `path`. Any
-#'   name obtained from `path` overrides this argument. Defaults to current
-#'   name.
+#' @param name Character, new file name if not specified as part of `path`. This
+#'   will force `path` to be treated as a folder, even if it is character and lacks
+#'   a trailing slash. Defaults to current name.
 #' @template verbose
```
jennybc: Ditto the above comment. Template?

R/drive_mv.R
```diff
@@ -52,8 +52,13 @@ drive_mv <- function(file, path = NULL, name = NULL, verbose = TRUE) {
...
-    if (is.null(name) && drive_path_exists(append_slash(path))) {
-      path <- append_slash(path)
+    if (is.null(name) && !has_slash(path) && drive_path_exists(append_slash(path))) {
+      stop(
+        "Unclear if `path` specifies parent folder or full path\n",
+        "to the new file, including its name. ",
+        "See ?as_dribble() for details.",
+        call. = FALSE
+      )
     }
```
jennybc: It feels like we should have a helper and reuse it in multiple places. For the check or for the message.

R/drive_publish.R
```diff
@@ -64,13 +74,16 @@ drive_change_publish <- function(file,
       message(
         glue_data(
           file_update,
-          "You have changed the publication status of {sq(name)}."
+          "\nYou have changed the publication status of file:\n",
+          "  * {sq(name)}."
         )
```
jennybc: When we only pull one value out of something like `file_update`, perhaps we should shift to `glue("... {file_update$name}")`.

R/drive_share.R
```diff
@@ -17,6 +17,21 @@
...
+#'    type = "document"
+#'    )
+#'
+#' ## Share file
+#' file %>%
+#'   drive_share(role = "reader", type = "anyone")
+#'
+#' ## Clean up
+#' drive_rm(file)
+#' }
```
jennybc: I applaud the addition of examples and self-cleaning-up examples 👏

R/drive_upload.R
```diff
@@ -8,8 +8,9 @@
 #'
 #' @template media
 #' @template path
-#' @param name Character, name the file should have on Google Drive if not
-#'   specified in `path`. Will default to its local name.
+#' @param name Character. The name the file should have on Drive. This will force
+#'   `path` to be treated as a folder, even if it is character and lacks
+#'   a trailing slash. Will default to its local name.
 #' @param type Character. If `type = NULL`, a MIME type is automatically
```
jennybc: yada yada

R/drive_upload.R
```diff
@@ -45,16 +46,21 @@ drive_upload <- function(media,
...
-    if (is.null(name) && drive_path_exists(append_slash(path))) {
-      path <- append_slash(path)
+    if (is.null(name) && !has_slash(path) && drive_path_exists(append_slash(path))) {
+      stop(
+        "Unclear if `path` specifies parent folder or full path\n",
+        "to the new file, including its name. ",
+        "See ?as_dribble() for details.",
+        call. = FALSE
+      )
     }
```
jennybc: yada yada

R/drive_upload.R
```diff
@@ -65,8 +71,17 @@ drive_upload <- function(media,
...
+    paths <- glue_data(path, "  * {name}: {id}")
+    stop(
+      collapse(
+        c("Requested parent folder identifies multiple files:", paths),
+        sep = "\n"
+      ),
+      call. = FALSE
+    )
+  }
   if (!is_folder(path)) {
```
jennybc: yada yada 😄

R/utils-paths.R
```diff
@@ -16,10 +16,15 @@ rootize_path <- function(path) {
   sub("^~$|^/", "~/", path)
 }
 
+## does path have a trailing slash?
+has_slash <- function(path) {
+  grepl("/$", path)
+}
+
```
jennybc: good idea

tests/testthat/test-drive_cp.R
```diff
@@ -86,3 +93,29 @@ test_that("drive_cp() errors if asked to copy a folder", {
...
+    file_cp <- drive_cp(nm_("i-am-a-file"), nm_("file-name"), nm_("file-name")),
+    "Requested parent folder does not exist"
+  )
+
+  ## if `path` identifies multiple files, it will error
+  expect_error(
+    file_cp <- drive_cp(nm_("i-am-a-file"), paste0(nm_("not-unique-folder"), "/")),
+    "Requested parent folder identifies multiple files"
+  )
+})
```
jennybc: Good to see more tests of these tricky cases.

jennybc:
I made some inline comments, but this is good progress on those issues. Note how I edited the first comment? That makes it easy to click and read the issues we're talking about. Before this is merged, you should edit the issue title to have "closes" or "fixes" in front of each issue number.

To really close all those issues, I think we need to find more of the common stuff and write a few helpers. By this I mean, some common checkers/messagers and/or a revival of functions like `spf()` that combine `stop()` or `message()` with some recurring glue patterns we use.

But what you've done here is necessary first step.

R/drive_mkdir.R
LucyMcGowan: I was thinking in this case it couldn't contain 0 files since we have this `path <- path %||% root_folder()`, so if it were empty, it would default to the root folder.

jennybc: Ah ok. This is probably good time to do yet another consistency check across `drive_cp()`, `drive_mv()`, `drive_mkdir()`. Would it simplify some of those other functions to do same, i.e. default to parent being root folder?

LucyMcGowan: For some reason to me (at the moment) `drive_mkdir()` it makes sense to default to making a directory in your root directory, but for `drive_mv()` if you don't give a good place for it to move to, I'm not sure that the root directory makes the most sense? And for `drive_cp()`, I think maybe the default should actually be sticking it in the folder the original file is in rather than ow we have it error now. What do you think?

LucyMcGowan:
@jennybc I've made a few updates. I ended up removing the individual glue function imports, since we are mostly using the helper functions now (but if there is a reason to put them back in I can change it back!)

I changed a few of the defaults for folder location if not given - I'm happy to change those though.

R/drive_mkdir.R
jennybc: Re: parent folder.

`drive_cp()`: I now remember saying something along these lines: "if user doesn't give instructions about parent folder, we should not even mention it in the request". I still feel this way. So I think we leave it alone.

`drive_mv()` and `drive_mkdir()`: what you say makes sense.

R/utils-paths.R
```diff
@@ -85,3 +90,14 @@ apply_extension <- function(path, ext) {
...
 }
+
+confirm_clear_path <- function(path, name) {
+  if (is.null(name) && !has_slash(path) && drive_path_exists(append_slash(path))) {
+    stop(
+      "Unclear if `path` specifies parent folder or full path\n",
+      "to the new file, including its name. ",
+      "See ?as_dribble() for details.",
+      call. = FALSE
+    )
```
jennybc: Should this be using `sglue()`?

R/drive_cp.R
```diff
@@ -47,35 +49,39 @@ drive_cp <- function(file, path = NULL, name = NULL, verbose = TRUE) {
   }
 
   if (is_path(path)) {
-    if (is.null(name) && drive_path_exists(append_slash(path))) {
-      path <- append_slash(path)
-    }
+    confirm_clear_path(path, name)
```
jennybc: I don't love this name, but neither can I suggest a better one right now. It's internal anyway so 👍

R/drive_cp.R
```diff
@@ -47,35 +49,39 @@ drive_cp <- function(file, path = NULL, name = NULL, verbose = TRUE) {
...
+      c("Requested parent folder identifies multiple files:", paths),
+      sep = "\n"
+    )
+  }
+  ## if path was input as a dribble or id, still need to be sure it's a folder
+  if (!is_folder(path)) {
+    sglue("\n`path` specifies a file that is not a folder:\n * {path$name}")
+  }
+  params[["parents"]] <- list(path$id)
+
```
jennybc: I'm not sure about this (the copying of the parent). Can you check against Drive operations in the browser? If file A has folder B as parent and you make a copy of A, does the copy have folder B as parent?

It feels like we should either stay out of this or do some global treatment of metadata. But mostly it feels like something we shouldn't get into. If we do it, what about multiple parents? Because that doesn't seem accounted for.

R/drive_mime_type.R
```diff
@@ -58,11 +58,9 @@ drive_mime_type <- function(type = NULL) {
   mime_type <- .drive$mime_tbl$mime_type[m]
 
   if (all(is.na(mime_type))) {
-    stop(glue(
-      "Unrecognized `type`:\n{problems}",
-      problems = collapse(type[is.na(mime_type)], sep = "\n")
-    ),
-    call. = FALSE
+    sglue(
+      "\n Unrecognized `type`:\n  * {problems}",
```
jennybc: extra space between `"\n"` and `Unrecognized`

R/googledrive-package.r
```diff
@@ -3,5 +3,4 @@
 #' @name googledrive
 #' @docType package
 #' @importFrom purrr %||%
-#' @importFrom glue glue glue_data collapse
```
jennybc: I lean towards keeping these imports.

R/utils.R
```diff
@@ -2,6 +2,17 @@ last <- function(x) x[length(x)]
 
 sq <- function(x) glue::single_quote(x)
 
+sglue <- function(...) stop(glue::glue(..., .envir = parent.frame()), call. = FALSE)
```
jennybc: Are you doing `glue(..., .envir = parent.frame())` instead of `glue(...)` for a specific reason? I would have expected the latter.

R/utils.R
```diff
@@ -2,6 +2,17 @@ last <- function(x) x[length(x)]
...
+sglue_data <- function(...) stop(glue::glue_data(..., .envir = parent.frame()), call. = FALSE)
+scollapse <- function(...) stop(glue::collapse(...), call. = FALSE)
```
jennybc: Do we make any collapse calls where `sep` is NOT `"\n"`? It feels prevalent enough that perhaps our wrapper should default to that.

man-roxygen/name.R
```diff
@@ -0,0 +1,3 @@
+#' @param name Character, new <%= name %>  name if not specified as part of `path`. This
```
jennybc: Is there one extra space here? `<%= name %>  name`

R/utils.R
LucyMcGowan: For some reason I get weird behavior if I don't include the `.envir = parent.frame()`. If you run this does it work?
```
mglue <- function(...) message(glue::glue(...))
x <- function(val) mglue(val)
x("test")
```
On my machine it can't find `val` unless I explicitly set the `.envir`.

jennybc:
Looking good!

How strongly to you feel about the names like `scollapse()`, etc.? My instinct is to be more obvious with `stop_collapse()`, `message_glue()`, etc.

LucyMcGowan:
I don't feel strongly at all! I'll update to your suggestion :+1:

R/utils.R
jennybc: The alternative is to do what you see here: https://github.com/tidyverse/glue/issues/7. That's the issue where, one day, glue might actually offer a replacement for our DIY versions.

jennybc: Do as you see in the glue issue. BTW, at that time it was written `f()` is what we now call `glue()`.

LucyMcGowan: 👌 will do. He also sets the `.envir` there - it seems strange that you have to but 🤷‍♀️ .

jennybc: I think it's because, if you wrap `glue()`, you need to preserve and convey the correct environment for the template variables. `.envir` is not a simple, static default value.

jennybc:
Thanks!

PR #153, opened 2017-07-25
Basic q handling in drive_find(); closes #38

### LINKED ISSUE #38: Helpers for working with q in drive_find()
We need some internal support for q handling. I'll come back to flesh this out. Brain dump and notes from #7.

> Forming q. At its simplest, people have to provide q verbatim. But it might be nice to accept a vector of q's or multiple q's in ..., which `drive_list()` would then concatenate with "and". It feels like there's also a need for an internal "append this to q" helper.

How do we -- or do we -- reconcile defaults (such as “not being in trash”) with user’s supplied q and/or q clauses we build up? Re: user … should they be forced to declare whether their q should trump everything or be added to the defaults?

Do we attempt to pull user-supplied q apart into clauses? Have some clause-checking machinery, then “and” it all back together?

jennybc:
@LucyMcGowan Would you find a couple ways to use this and sanity check? I've tried it and there's tests, but there's no substitute for another human 🙂.

### DIFF
--- a/R/drive_find.R
+++ b/R/drive_find.R
@@ -1,9 +1,19 @@
 #' Find files on Google Drive.
 #'
-#' This is the closest googledrive function to what you get from
-#' <https://drive.google.com>: by default, you just get a listing of your files.
-#' You can also narrow the search in various ways, such as by file type, whether
-#' it's yours or shared with you, starred status, etc.
+#' @description This is the closest googledrive function to what you get from
+#'   <https://drive.google.com>: by default, you just get a listing of your
+#'   files. You can also narrow the search in various ways, such as by file
+#'   type, whether it's yours or shared with you, starred status, etc.
+#'
+#' By default, `drive_find()` does not show files in the trash: it adds
+#'   `q = "trashed = false"` to the query. However, it will not do so if the
+#'   user specifies a `q` search clause for trash inclusion or exclusion. To see
+#'   only files in the trash, use [drive_view_trash()], which is a shortcut for
+#'   `drive_find(q = "trashed = true")`. To see files regardless of trash
+#'   status, use `drive_find(q = "trashed = true or trashed = false")`.
+#'
+#' `drive_find()` will accept multiple `q` clauses and/or a vector `q` of
+#'   several search clauses. These clauses are combined with `and`.
 
 #' @seealso Helpful links for forming queries:
 #'   * <https://developers.google.com/drive/v3/web/search-parameters>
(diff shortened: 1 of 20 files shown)


R/drive_find.R
```diff
@@ -64,25 +86,17 @@ drive_find <- function(pattern = NULL,
 
   params <- list(...)
   params$fields <- params$fields %||% prep_fields(drive_fields())
+  params <- marshal_q_clauses(params)
```
LucyMcGowan: 😍 I LOVE this name so much!

LucyMcGowan:
👌 It looks great. I played a bit - I basically just tried multiple combinations of the terms listed in the [API Docs](https://developers.google.com/drive/v3/web/search-parameters). 

One thing I found (maybe this should be this way, but just thought I'd point it out). When I misspecify a `q` term, for example forget to put a name in single quotes, the error is from `prossess_response()`. I wonder if we should be proactive about that? 
```r
drive_find(q = "name contains unconftest")
##> Error: HTTP error [400] Invalid Value
##>   * domain: global
##>   * reason: invalid
##>   * message: Invalid Value
##>   * locationType: parameter
##>   * location: q
##> Called from: process_response(page)
```

jennybc:
That would be nice but, I think, quite a lot of work. If you see a simple way to do that, speak up. But AFAICT, that falls in the realm of writing a mini-DSL for `q`. Which sounds like fun but is best left for later, I think. I have various other ideas too, like accepting straight up logicals, but I think it would be several days of work. Let's look on the bright side: at least the error says it's about `q`! 😄

LucyMcGowan:
That's true! Should we add in `call. = FALSE` to that [process_request() error](https://github.com/tidyverse/googledrive/blob/master/R/process_response.R#L36) (Sorry if that is a silly question 🙈 still trying to work out the tidyverseisms)

jennybc:
It seems to create a double error message if I do. I'm not very experienced creating custom error classes, as I have done here and think is a good idea long-term, so I'll just leave it.

jennybc:
Without out `call. = FALSE` -- so the way I wrote it -- I do not see what you see. Here's what I get:

```
> drive_find(q = "name contains unconftest")
Error: HTTP error [400] Invalid Value
  * domain: global
  * reason: invalid
  * message: Invalid Value
  * locationType: parameter
  * location: q 
```

Wonder what the difference is?

LucyMcGowan:
🤔 Perhaps it is just because I am in dev mode? It shows when I run in my console, but not when I load the package fresh & run elsewhere (so I suppose it doesn't matter)

Issue #152, opened 2017-07-22
What metadata to copy with drive_cp()?


jennybc:
Parent folder(s) is a specific one we've discussed a lot, but maybe there's other metadata that should also be considered.

I note that using Drive in the browser ... a copy appears in the same folder as the original. I haven't looked at the multiple parents situation specifically.

jennybc:
One key difference between our situation and Drive in the browser: in the browser, you are implicitly "in" a folder, either the root folder or something else. So it makes sense to make the copy there.

If you are in a Google Doc (or, I imagine, Sheet or Presentation) and you do *File > Make a copy*, you get this:

<img width="342" alt="screen shot 2017-07-24 at 10 58 52 am" src="https://user-images.githubusercontent.com/599454/28537349-5943e574-705f-11e7-92a6-bacb59227d44.png">

i.e. it defaults to the (a?) existing parent of the file, but allows you say otherwise. I note that the options for specifying a different folder are weird/awkard, but that's neither here nor there.

So far, I conclude we *should* default to copying a file's parent folder. And if it has more than one ... I guess we have to pick one?

jennybc:
Note also the option above to transfer sharing settings, which is relevant to the larger question of which metadata to copy.

LucyMcGowan:
I've done a small study on this in a few circumstances.

It seems if a file has multiple parents it defaults to copying it into the _current folder_, not the (potentially multiple) files - so maybe we can default to copying into the first parent if not specified?

Regarding sharing settings, Drive does not copy sharing setting in the UI, so I think it's okay for us to do the same.

What do you think?

jennybc:
I didn't formally close this when I merged #158. Now that `drive_cp()` accepts `...`, is it possible we should do nothing here? That is, accept whatever the API does by default and if the user wants to transfer metadata, they have to do it themselves? It can be done right now, in several different ways, but it's clunky.

I don't think googledrive should do anything re: general metadata transfer by default. But we could offer an interface in `drive_cp()` for specifying metadata properties that should be copied from source to target.

PR #158, opened 2017-07-28
Update handling of metadata in update and copy requests (fixes #127, fixes #148)

### LINKED ISSUE #127: drive_copy() needs to expose ...
https://developers.google.com/drive/v3/reference/files/copy

For discussion: It should be possible to send those optional query parameters and to patch more of metadata via the body.

### LINKED ISSUE #148: Generalize drive_update()
We will soon have `drive_update()` once #141 gets merged. 🎉

Right now it's hard-wired to do a media upload. What if you want to update metadata (alone or along with media update)?

LucyMcGowan:
Generalize drive_update() #148 

drive_copy() needs to expose ... #127

Related but so far not resolved in this PR:

What metadata to copy with drive_cp()? #152


DESCRIPTION
```diff
@@ -17,7 +17,8 @@ Imports: glue (>= 1.1.0),
   httr,
   purrr,
   tibble,
-  utils
+  utils,
+    jsonlite
```
jennybc: I like to keep this section alphabetical and uniformly indented. 🤓

R/drive_cp.R
```diff
@@ -11,6 +11,7 @@
 #' @template name
 #' @templateVar name file
 #' @templateVar default Defaults to "Copy of `FILE-NAME`".
+#' @param ... Parameters to pass along to the API query.
```
jennybc: Should say a bit more about what this might be? I.e. whatever you might accomplish re: metadata with `...`.

jennybc: I think there should also be at least one example that uses `...`. For our own future selves, if nothing else.

R/drive_update.R
```diff
@@ -6,7 +6,8 @@
 #'
 #' @template file
 #' @template media
-#' @param ... Parameters to pass along to the API query. NOT IMPLEMENTED YET.
+#' @param ... Parameters to pass along to the [API query](https://developers.google.com/drive/v3/reference/files/update).
+#'   These parameters can be used to update metadata.
```
jennybc: This is what I want for `drive_cp()` documentation of `...`. In fact, could they share a template? You can move the endpoint-specific hyperlink into another part of the help file.

R/drive_update.R
```diff
@@ -41,24 +38,85 @@ drive_update <- function(file,
     stop_collapse(c("Path to update is not unique:", files))
   }
 
+  if (!is_mine(file)) {
+    stop_glue("Can't update this file because you don't own it:\n{file$name}")
```
jennybc: bulletize

R/drive_update.R
```diff
@@ -41,24 +38,85 @@ drive_update <- function(file,
...
+  }
+
+  params <- list(
+    fileId = file$id,
+    fields = "*",
+    ...)
+
+  if (length(params) == 2L) {
```
jennybc: I would sleep better if this consulted actual names vs. just length.

jennybc:
I've done a fairly superficial review but maybe you could respond to those comments ... and then I'll do a better review later today.

R/drive_cp.R
```diff
@@ -11,12 +11,13 @@
 #' @template name
 #' @templateVar name file
 #' @templateVar default Defaults to "Copy of `FILE-NAME`".
+#' @template dots
```
jennybc: Suspect this should be something more specific, like `dots-metadata`. First, someone might misuse it if they think it's a general dots documentation. Second, we might very well have more than one template re: dots before this all over.

R/drive_mv.R
```diff
@@ -87,11 +93,8 @@ drive_mv <- function(file, path = NULL, name = NULL, verbose = TRUE) {
     }
   }
 
-  request <- generate_request(
```
jennybc: Comment about this line above (I can't comment directly because it's not part of this diff):

```
current_parents <- file$files_resource[[1]][["parents"]][[1]]
```

this is not going to get the actual "current parents" in the case of multiple parents. It gets just one. We need to think about this case. If there are multiple parents ... maybe we only add the new one and we message about this? My logic is that this person / file has already crossed the Rubicon into multiple parents, so why not one more? If there is exactly one parent, which will be the case 99% of the time, we do the conventional add + remove.

jennybc: Note, this will affect the logic for the messaging re: renamed vs moved vs both.

R/drive_mv.R
```diff
@@ -38,12 +44,20 @@
 #' ## specify destination as path with trailing slash
 #' ## to ensure we get a move vs. renaming it to "new-folder"
 #' file <- drive_mv(file, "new-folder/")
+#'
+#' ## TO DO: clean up
```
jennybc: Can you add the clean up here?

R/drive_update.R
```diff
@@ -6,20 +6,30 @@
 #'
 #' @template file
 #' @template media
-#' @param ... Parameters to pass along to the API query. NOT IMPLEMENTED YET.
```
jennybc: The description will need to be expanded here. Commenting on this line, just because I can.

R/drive_update.R
```diff
@@ -41,24 +47,85 @@ drive_update <- function(file,
...
+    stop_glue("\nCan't update this file because you don't own it:\n  * {file$name}")
+  }
+
+  params <- list(
+    fileId = file$id,
+    fields = "*",
+    ...)
+
+  if (identical(names(params), c("fileId", "fields"))) {
+    response <- drive_update_media(file = file, media = media)
```
jennybc: I think we need to either:

  1. Push `as_dribble(list(process_response(...)))` down into these 3 functions. *I think this is my preference?*
  1. Rename them to something like `do_update_*()`.

because right now their names suggest they are like our other `drive_*()` functions, i.e. that they return a dribble. Yet they do not.

jennybc:
Something about the multipart upload is sensitive to new changes in httr. I'm not sure if you need to write it differently or if there's a regression in httr and you should create a reprex and open an issue there. You will see what I mean if you test `drive_update()` with CRAN httr vs current HEAD on GitHub.

jennybc:
I plan to list curl in Imports, btw, and require version >= 2.8.1, in case that becomes relevant.

R/drive_mv.R
jennybc: Showing the addition or deletion of a parent would make a great example for `drive_update()`, actually. It's a legit example, plus people might need to do this in multiple parent situations and we provide no direct support for that. You should also rip out my un-exported `drive_add_parent()` function and replace its usage in test file setup with a call to `drive_update()`.

LucyMcGowan:
I think I downloaded the dev httr again and didn't notice an issue (I currently have `1.2.1.9000` - what was the error you saw with tests?

jennybc:
I saw this while fooling around with this PR:

```
> make_request(request)
 Error in UseMethod("as_field") : 
  no applicable method for 'as_field' applied to an object of class "form_file" 
```

I'm not sure exactly which dev version of httr I had at the time. I have now reverted to CRAN, to see if I could pass tests (I did). So maybe I was at some awkward intermediate commit and the current HEAD is ok for us. Or perhaps I was just in some peculiar state. Glad to hear all seems well.

R/drive_cp.R
```diff
@@ -57,9 +66,11 @@ drive_cp <- function(file, path = NULL, name = NULL, verbose = TRUE) {
...
   name <- name %||% glue("Copy of {file$name}")
 
-  params <- list(
+  dots <- list(...)
+  dots$fields <- dots$fields %||% "*"
+  params <- c(
     fileId = file$id,
-    fields = "*"
+    dots
   )
```
jennybc: This is how I propose we deal with the possibility that user passed `fields` in `...`. I also now think we should default to `fields = "*"` everywhere and see if there's ever real world justification for working any harder at this.

PR #167, opened 2017-08-02
Allow metadata to pass through drive_upload (closes #163)

### LINKED ISSUE #163: drive_upload() and metadata
Once #158 is done, we should revisit `drive_upload()`. It should probably have greater capabilities re: metadata.

LucyMcGowan:


R/drive_upload.R
```diff
@@ -40,11 +41,33 @@
...
+#' ## Create mutltiple parent directories
+#' folder_1 <- drive_mkdir("folder_one")
+#' folder_2 <- drive_mkdir("folder_two")
+#'
+#' ## Upload the file into multiple parents. Notice here we are using
+#' ## the `addParents` parameter via the `...`.
+#' logo <- drive_upload(
+#'   R.home('doc/html/logo.jpg'),
+#'   folder_1,
+#'   addParents = folder_2$id
```
jennybc: FWIW I suspect you can both at the same time if you do `addParents = paste(folder_1$id, folder_2$id, sep = ",")`.

But, in any case, I'd prefer a less "out there" example. Such as adding a description? Or starring the file? I don't want to encourage anyone to do multiple parents!

R/drive_upload.R
```diff
@@ -40,11 +41,33 @@
...
+#' )
+#'
+#' ## Examine the multiple parents
+#' drive_add_path(logo)
+#'
+#' ## Clean up
+#' bind_rows(folder_1, folder_2, logo) %>%
```
jennybc: You'll need to just do three separate `drive_rm()` calls. We don't want to use dplyr in our examples. It's only a suggested package. You could use a naming scheme that allows you to pipe the result of `drive_find()` into `drive_rm()`.

tests/testthat/test-drive_upload.R
```diff
@@ -38,3 +38,40 @@ test_that("drive_upload() places file in non-root folder, with new name", {
...
+})
+
+test_that("drive_upload() accepts query metadata via ...", {
+  skip_if_no_token()
+  skip_if_offline()
+  on.exit(drive_rm(me_("DESCRIPTION")))
+
+  ## To ensure query metadata gets passed, we are going to add this
+  ## parent, so the file should be both in My Drive and "upload-into-me"
+  destination <- drive_get(nm_("upload-into-me"))
```
jennybc: Hmmm ... so does this imply that every uploaded file will have root folder as parent? I.e. we can never upload such that the only parent is some other folder? Just curious what you know.

R/drive_upload.R
jennybc: Now that I've read the tests, I understand this better now. Your approach prevents root folder from being a parent, IIUC.

tests/testthat/test-drive_upload.R
jennybc: Nevermore. You can ignore the above question.

jennybc:
Should `drive_upload() handle parent folder just like `drive_cp()` now does? If user doesn't provide a specific parent, it doesn't seem our place to do so either. I acknowledge that I'm probably the one who made it this way to begin with. Since then, we seem to have developed a policy of never enforcing default params unless we absolutely must. Seems like a good idea.

jennybc:
Two more questions:

  * Should we create and export `drive_create()` for creation of a file id but meta-data only? I lean towards "no", but seems worth asking. One argument in favour is that, once I handle Team Drives, there will be a `team_drive_create()` function but no companion `drive_create()` for a regular Drive file. Then again `team_drive_create()` is arguably more like `drive_mkdir()`, which does exist.
  * Each `drive_upload()` call currently hits the API twice: once to create the file id and again to upload media. Can we do it all at once with a multipart request?

jennybc:
The endpoint added in https://github.com/tidyverse/googledrive/pull/168 is presumably relevant to this PR, if you make this function more like `drive_update()` wrt multipart requests.

LucyMcGowan:
@jennybc I updated `drive_upload()` to only make a single API call - in doing this, we no longer can add or remove parents (since those parameters are not available for the `drive.files.create` endpoint, only for `drive.files.update`) - this is not really a problem, just wanted to give you the heads up so you knew why I changed the tests.

jennybc:
It possible to control the parents (plural) at create time:

> `parents[]` The IDs of the parent folders which contain the file. If not specified as part of a create request, the file will be placed directly in the My Drive folder. Update requests must use the `addParents` and `removeParents` parameters to modify the values.

which, of course, we know, since we use that property ourself if `path` specifies a parent folder. So if `path` specifies a parent, I think you need to check if user's `...` includes `parents` and send the union. Do you understand and agree?

jennybc:
Also, just checking: an media-only upload with no metadata still works, yes? No parent, no nothing?

LucyMcGowan:
The media-only uploads do work, but as I have it now they go through the same multipart process as any other upload (do you think this is okay? Right now it's a tad expensive because we're saving a temp file, but once httr updates, it should be pretty straightforward). I'm working on some `parent` & `path` tests/errors this morning 🌻.

jennybc:
> they go through the same multipart process as any other upload (do you think this is okay

Could this function follow the pattern of `drive_update()`? If it can, it should. It will be less baffling to us in future.

LucyMcGowan:
It seems a bit difference because they'd either be uploading media only or media + metadata - also I realized I misspoke a bit above (I've edited) because you actually will always have a `name` since we either take the given name or base name of the media.

LucyMcGowan:
🎉 Yes I think so!

PR #169, opened 2017-08-03
Support Team Drives; fixes #101

### LINKED ISSUE #101: Team Drives
We have basically decided to NOT pass parameters through willy nilly in our high level functions. But we have to provide a global way to accommodate files on Team drives. I would prefer to think about having some default query params that are sent automatically for relevant endpoints + a way to manipulate them vs. putting `...` everywhere.

https://developers.google.com/drive/v3/web/enable-teamdrives

Good news:

> If your application isn't affected by the behavioral differences, then you only need to include the `supportsTeamDrives=true` query parameter in requests. This is an acknowledgement that the application is aware of behavioral differences of files contained in Team Drives.

Bad news:

`files.list` endpoint obviously has some special considerations re: Team Drives. Ditto for permissions.

jennybc:
Fixes #101 

At least enough for people to accumulate some usage. Still need to expand the testing framework to exercise this formally.


jennybc:
About to merge this @LucyMcGowan. It's enormous and yet ... if Team Drive aren't available/relevant to you ... it should do absolutely nothing 😑. But definitely take note and have a quick skim. If you have any questions re: implications, let's discuss in our meeting.

PR #173, opened 2017-08-10
Permissions overhaul closes #172, closes #129, closes #171

### LINKED ISSUE #129: How do you examine current sharing status?
@LucyMcGowan How do you examine current sharing status? Seems like we don't have a nice workflow for that.

### LINKED ISSUE #171: Review docs for clarity re: what's vectorized and what's not
Example of how lack of clarity (or, really, vigorous use of templates) can make it harder for a user to figure out what's going on: #170.

@LucyMcGowan will you take a look at the implicated template(s) and do something to fix? And any adjacent parameter templates? If you accept, you can just assign this to yourself. 🙂

### LINKED ISSUE #172: Naming pattern for functions that reveal details
We need to do something to unify functions that show you the current state of files in a dribble wrt trashed-ness, path, sharing status, publishing status, etc.

Should it be:

`drive_show_trash()`
`drive_show_path()`
`drive_show_sharing()`
`drive_show_publish()`
etc.
?

Should be clearly distinct from functions that **do** something like put files in the trash, move files, or alter the sharing/publishing status.

LucyMcGowan:
How do you examine current sharing status? #129

Naming pattern for functions that reveal details #172

Review docs for clarity re: what's vectorized and what's not #171


R/drive_find.R
```diff
@@ -26,7 +26,7 @@
 #' By default, `drive_find()` does not include files in the trash: it adds `q =
 #' "trashed = false"` to the query. However, it will not do so if the user
 #' specifies a `q` search clause for trash inclusion or exclusion. To see only
-#' files in the trash, use [drive_view_trash()], which is a shortcut for
+#' files in the trash, use [drive_show_trash()], which is a shortcut for
```
jennybc: Overall, this is clearly going to be a great step forward re: permissions and featuring other info we excavate from `drive_resource`.

One nagging concern re: trash is that this is a row-based operation. We get the trash status, filter on it, and don't touch the columns. The path/publish/sharing versions are consistent: we feature some info in one or more new columns. 🤔

LucyMcGowan: ah that's a good point - would it be preferable to have `drive_show_trash()` add a logical `trash` column (but not filter on the rows), and then leave `drive_view_trash()` to filter for only those in the trash?

LucyMcGowan:
@jennybc If you wouldn't mind 👀 at the `drive_show_share()` and friends - I am not sure I :heart: how I do it right now. Essentially you provide both the `role` and what I am calling `display` and it will pull the information into a single cell (for example, if you have multiple readers, and you set `display = "name"` you'd add a column `reader` to your `dribble` with comma separated names for all your readers).

R/drive_publish.R
```diff
@@ -40,21 +40,38 @@ drive_change_publish <- function(file,
                                  publish = TRUE,
                                  ...,
                                  verbose = TRUE) {
-  file_update <- drive_is_published(file = file, verbose = FALSE)
-
-  file_update <- confirm_single_file(file_update)
+  file_update <- drive_show_publish(file = file, verbose = FALSE)
+  file_update <- confirm_some_files(file_update)
+  file_update <- split(file_update, 1:nrow(file_update))
```
jennybc: Do you really need to `split()`? Might be a place to use `pmap()`. Or just `map()` on `file$drive_resource`? Consider also if the elevation of publishing info belongs here or down in `change_publish_one()`. If `change_publish_one()` returned one drive file, then a map over files returns a list of drive files, and we can `as_dribble()` it, as done elsewhere. Main point: use a mapping pattern we have elsewhere if can.

R/drive_publish.R
```diff
@@ -132,24 +135,24 @@ drive_is_published <- function(file, verbose = TRUE) {
...
       "Check out `drive_share()` to change sharing permissions."
     ))
   }
 
-  published <- purrr::map2(file$id, file$name, is_published_one, verbose = verbose)
-  file$publish <- published
-
+  files <- split(file, 1:nrow(file))
+  files <- purrr::map(files, show_publish_one, verbose = verbose)
+  file <- do.call(rbind, files)
```
jennybc: Same comment as above re: looking around for existing pattern.

R/drive_publish.R
LucyMcGowan: Ah probably `pmap()`, or `map()` on `drive_resource` would be better. ⚠️ you are going to see `split()` a few times since I tried to stay somewhat consistent, but I will systematically change those on my next push.

R/drive_publish.R
```diff
@@ -158,18 +161,26 @@ is_published_one <- function(id, name, verbose = TRUE) {
...
-    revision = x$id,
-    published = x$published,
-    auto_publish = x$publishAuto %||% FALSE
+add_publish_cols <- function(d, x) {
+  if ("published" %in% names(d)) {
+    d$published_check_time <- Sys.time()
+    d$revision_id <- x$id
+    d$published <- x$published
+    d$auto_publish <- x$publishAuto %||% FALSE
+    return(d)
```
jennybc: I would order these by human interest. So `published` should come first, for example!

jennybc:
Why don't you ping me when you're at another good stopping point and I will actually pull vs. doing a drive-by review.

LucyMcGowan:
👌 @jennybc I think this is ready for a look

jennybc:
I'm going to build up a proposal on how to finish this off here. Will signal when I'm done.

Let's create `drive_reveal()`. It adds one or more columns to an existing dribble. It never changes the number of rows. It adds the columns right after `name`. It has an argument, maybe `what`?, that takes on one or more values: path, trashed, sharing, publishing.

jennybc:
A few more concrete things:

Add `@seealso` re: <https://developers.google.com/drive/v3/reference/revisions/update>

Perhaps introduce `auto` as an argument to `drive_publish()` that ultimately feeds into `publishAuto`

NAMESPACE
```diff
@@ -44,16 +43,17 @@ export(drive_extension)
...
 export(drive_ls)
 export(drive_mime_type)
 export(drive_mkdir)
 export(drive_mv)
 export(drive_publish)
 export(drive_rename)
+export(drive_reveal)
 export(drive_rm)
 export(drive_share)
+export(drive_show_trash)
```
jennybc: Is it intentional to keep exporting this?

R/drive_get.R
```diff
@@ -6,7 +6,7 @@
 #'
 #' @description If the files are specified via `path`, the returned [`dribble`]
 #'   will include a `path` variable. To add path information to any [`dribble`]
-#'   that lacks it, use [drive_add_path()]. If you want to list the contents of
+#'   that lacks it, use [drive_reveal()]. If you want to list the contents of
 #'   a folder, use [drive_ls()]. For general searching, use [drive_find()].
```
jennybc: `drive_reveal(what = "path")`?

I will probably dig in a bit here and, if I do, I'll make this change. Or you can and I'll pull.

R/drive_mime_type.R
```diff
@@ -105,3 +105,12 @@ one_ext <- function(type) {
   }
   m
 }
+
+drive_show_mime_type <- function(file) {
+  file <- as_dribble(file)
+  file <- confirm_some_files(file)
+  file[["mime_type"]] <- NULL
```
jennybc: What's the motivation for this? I was pretty careful in `promote()` re: edge cases, so am curious if I missed one.

R/drive_publish.R
```diff
@@ -1,9 +1,15 @@
-#' Publish Google Drive file.
+#' Publish Google Drive file
```
jennybc: I think we've decided to stop saying "Google Drive" and, even, "Drive" so much, right? I think we should qualify only in very introductory materials or if we need to explicitly distinguish Drive files from local ones.

R/drive_publish.R
```diff
@@ -1,9 +1,15 @@
...
 #'
-#' @template file
+#' Publish a Google Drive file to the web. You can only publish "Google" files, such
```
jennybc: Elsewhere we call these "native" or "Google native" files, I think. Regardless, let's be consistent.

R/drive_publish.R
```diff
@@ -1,9 +1,15 @@
...
+#' as Google Documents, Google Spreadsheets, or Google Slides.
+#'
+#' @seealso Wraps the `revisions.update` endpoint:
+#'   * <https://developers.google.com/drive/v3/reference/revisions/update>
+#'
+#' @template file-plural
 #' @param ... Name-value pairs to add to the API request body, for example
 #'   `publishAuto = FALSE` will ensure that each subsequent revision will not be
-#'   automatically published (default here is `publishAuto = TRUE`).
+#'   automatically published (default here is `publishAuto = TRUE` and `publishedOutsideDomain = TRUE`).
```
jennybc: Let's make the example feature our default, i.e. amplify what it means for `publishAuto = TRUE`.

R/drive_publish.R
```diff
@@ -26,150 +32,119 @@ drive_publish <- function(file, ..., verbose = TRUE) {
 
 #' Unpublish Google Drive file
 #'
-#' @template file
-#' @param ... Name-value pairs to add to the API request body.
-#' @template verbose
+#' Unpublish a Google Drive file to the web. You can only publish "Google" files, such
+#' as Google Documents, Google Spreadsheets, or Google Slides, therefore you cannot
+#' (and there would be no need to) unpublish non-"Google" type files.
```
jennybc: `drive_publish()` and `drive_unpublish()` seem like good candidates to document together. Then you don't have to repeat this blurb, the examples are easier to write, etc.

R/drive_publish.R
```diff
@@ -26,150 +32,119 @@ drive_publish <- function(file, ..., verbose = TRUE) {
...
+      "Your file(s) are type:",
       all_mime_types,
       "Check out `drive_share()` to change sharing permissions."
     ))
   }
 
-  published <- purrr::map2(file$id, file$name, is_published_one, verbose = verbose)
-  file$publish <- published
-
+  params <- list(...)
```
jennybc: We need to `toCamel()` this. We're doing that everywhere we accept dots.

R/drive_publish.R
```diff
@@ -26,150 +32,119 @@ drive_publish <- function(file, ..., verbose = TRUE) {
...
+  params[["publishedOutsideDomain"]] <- params[["publishedOutsideDomain"]] %||% TRUE
+  params[["revisionId"]] <- "head"
+  params[["fields"]] <- "*"
+
+  revision_resource <- purrr::map(file$id,
+                                  change_publish_one,
+                                  params = params)
+  ## should we re-register the file here? technically when you publish,
+  ## the version increases by 1 (which is in drive_resource) so if
+  ## we don't re-register, the version will be incorrect.
```
jennybc: If you've verified the above is true, i.e. we will have the wrong revision number, then I lean towards "yes, get the metadata again". Or could you just patch that in? Is that the only thing known to be wrong with the metadata we have at that point?

R/drive_reveal.R
```diff
@@ -0,0 +1,40 @@
...
+#' ## Clean up
+#' drive_rm(file)
+#' }
+drive_reveal <- function(file, what = "path") {
+
+  reveal <- list("publish" = drive_show_publish,
+                 "permissions" = drive_show_permissions,
+                 "trash" = drive_show_trash,
+                 "path" = drive_show_path,
+                 "mime_type" = drive_show_mime_type)[[what]]
```
jennybc: What if `what` has length greater than one? Or doesn't match any of those? `file` also needs some validation.

R/drive_share.R
```diff
@@ -58,31 +65,59 @@ drive_share <- function(file, role = NULL, type = NULL, ..., verbose = TRUE) {
     )
   }
 
+  params <- list(...)
```
jennybc: `toCamel()`

R/drive_mime_type.R
LucyMcGowan: If you do promote 2x on the same file, I think it will just add an additional column instead of updating the current one.

R/drive_share.R
```diff
@@ -58,31 +65,59 @@ drive_share <- function(file, role = NULL, type = NULL, ..., verbose = TRUE) {
...
+  params[["role"]] <- role
+  params[["type"]] <- type
+  params[["fields"]] <- "*"
+  permissions_resource <- purrr::map(file$id,
+                                     drive_share_one,
+                                     params = params,
+                                     verbose = verbose)
+  file[["permissions_resource"]] <- NULL
+  file <- tibble::add_column(file, permissions_resource = permissions_resource, .after = 1)
```
jennybc: What have you experienced that makes you do this like so? Meaning: set `permissions_resource` to `NULL` then immediately overwrite it.

R/drive_publish.R
LucyMcGowan: I *think* I left it to be explicit (that only the Google type files work), but perhaps that can just go in the description?

R/drive_share.R
```diff
@@ -58,31 +65,59 @@ drive_share <- function(file, role = NULL, type = NULL, ..., verbose = TRUE) {
...
+  response <- make_request(request, encode = "json")
+  process_response(response)
+}
+
+drive_show_permissions <- function(file) {
+  file <- as_dribble(file)
+  file <- confirm_some_files(file)
+  permissions_resource <- purrr::map(file$id, show_permissions_one)
+  file[["permissions_resource"]] <- NULL
+  tibble::add_column(file, permissions_resource = permissions_resource, .after = 1)
```
jennybc: Same question as above.

R/drive_trash.R
```diff
@@ -64,14 +64,25 @@ toggle_trash_one <- function(id, trash = TRUE) {
...
-#' Empty Drive Trash.
+#' Add trash column to your dribble
+#' @template file-plural
+#' @template dribble-return
+#' @export
+drive_show_trash <- function(file) {
+  file <- as_dribble(file)
+  file <- confirm_some_files(file)
+  file[["trashed"]] <- NULL
+  promote(file, "trashed")
```
jennybc: Same Q.

R/drive_share.R
LucyMcGowan: If someone has already added "permissions_resource", `add_column()` will error.

NAMESPACE
LucyMcGowan: Nope 🙈

R/drive_mime_type.R
jennybc: Aha! Sounds like something I should address in `promote()`, instead of using this workaround everywhere, though. Agree?

LucyMcGowan: Ah yes! 🎉

R/drive_publish.R
jennybc: I see. But I don't think that's clear to most people. I think we can suppress it in places like titles. As commented elsewhere, we need to pick some vocabulary re: Docs, Sheets, etc. and then use it consistently. Yeah in the description, in this case.

R/drive_share.R
jennybc: So that would be true if ... they have a dribble that has passed through `drive_reveal()` already?

LucyMcGowan: yes - and also if they just shared something, since sharing alone adds a column (currently, but now that I've typed it I'm not certain that makes sense...)

jennybc: It is consistent with what happens when we identify files by path. Since we have path (and the user obviously likes to think about them), we include `path` as a column. It sounds sensible to include `permissions_resource` here, since you've got it, and the user is obviously thinking about permissions.

jennybc:
`drive_show_*()` functions: I think they should operate w/o error on a 0-row dribble, as `drive_show_path()` does. Right now the newer additions mostly use `confirm_some_files()`, but is there a specific reason why?

R/drive_mime_type.R
jennybc: I've done that now, but it doesn't help us here because of the `mimeType` vs `mime_type` issue. 😣

NAMESPACE
jennybc: fixed

R/drive_get.R
jennybc: fixed

R/drive_publish.R
jennybc: done

R/drive_publish.R
jennybc: done

R/drive_publish.R
jennybc: done

R/drive_publish.R
jennybc: done

jennybc:
`drive_reveal()`: besides general hardening (already commented), I want to discuss the values for `what`. If `what` is adding exactly one variable, perhaps its name should match the `what` value exactly?

R/drive_publish.R
jennybc: The `version` in the file resource is this:

> A monotonically increasing version number for the file. This reflects every change made to the file on the server, even those not visible to the user.

Based on an experiment I've done, this is NOT the same as the revision. And it doesn't increment by 1. I'm starting at a version of 7, even though the only thing I've done to the file is upload and publish. The revision is still at 1. But that's not exactly your point.

Perhaps we should call `drive_reveal(file, "publish")` and take care of it that way. Two good effects:

  * updates the metadata
  * reinforces modularity

jennybc: OK I did that ☝️

LucyMcGowan:
re: `drive_show_*()` I think I agree that it should be fine if there are no files, it just returns an empty dribble.

re: `drive_reveal()` the only one that returns more than one column is `publish` - it adds both the `published` column and the `revisions_resource` column. I added a quick check that the `what` value is valid.

R/drive_reveal.R
```diff
@@ -33,7 +33,16 @@ drive_reveal <- function(file, what = "path") {
...
+                 "mime_type" = drive_show_mime_type)
+
+  if (!(what %in% names(reveal))) {
+    stop_glue(
+      "\n'what' must be one of the following:\n",
+      "  * {collapse(names(reveal), sep = ', ')}."
+    )
+  }
+
+  reveal <- reveal[[what]]
```
jennybc: I think this will be cleaner overall if we put the vector of possible values as the default for `what`. Then inside the function you can use `match.arg()` to get `what`. And get the reveal function by using `switch(what, ...)`.

jennybc: done

PR #189, opened 2017-08-23
Add file.path() to examples that use R.home() so it works across platforms (closes #186)

### LINKED ISSUE #186: BioC_mirrors.csv not found
The examples in the "Upload files" section of the googledrive website (http://googledrive.tidyverse.org/#upload-files) uses the `BioC_mirrors.csv` file as an example, e.g. 

    (mirrors <- drive_upload(R.home('doc/BioC_mirrors.csv'), "README-mirrors.csv"))

For my R installation on Ubuntu 14.04, `R.home('doc/BioC_mirrors.csv')` returns `"/usr/lib/R/doc/BioC_mirrors.csv"`, yet there is no such file. However using

    `file.path(R.home("doc"), "BioC_mirrors.csv")`

as in `utils::chooseBioCmirror` returns `"/usr/share/R/doc/BioC_mirrors.csv"` which does exist, so specifying the file this way would fix this issue for me and presumably is a more general solution as that is what `utils::chooseBioCmirror` uses.

Note the path is used twice in this section.

LucyMcGowan:


jennybc:
FYI re: the second commit. For changes that only affect docs, broadly defined, you can include `[skip ci]` in the commit message and it won't do the appveyor/travis stuff.

jennybc:
Thanks!

PR #222, opened 2018-07-05
Support for upcoming glue 1.3.0 release

jimhester:
`glue::collapse()` has been renamed to `glue::glue_collapse()`, this
provides support for both versions.

This should let you release google drive whenever is convenient rather than relying on the glue release. However I do plan to submit glue 1.3.0 in the next few weeks.


jennybc:
I think this line in the tests is also implicated:

https://github.com/tidyverse/googledrive/blob/94c8c01cbb9fd96fe59920cf58e39263a3135337/tests/testthat/helper.R#L56

jennybc:
Thanks!

PR #258, opened 2019-06-07
Add function drive_create() for creating blank files

ianmcook:
I was recently using this package to create some blank Google Docs, like this:

```r
system("touch emptyfile")
drive_upload("emptyfile", type = "document", ... )
```
But I thought it'd be better if the package had a proper way to create blank files, so I implemented it here for your consideration. Thank you!


jennybc:
I'm curious: what do you use these empty documents for?

jennybc:
Thanks! I'm going to merge this and then massage it some more. In particular, `drive_mkdir()` just becomes a very thin wrapper around this.

jennybc:
I will add a NEWS bullet as well.

ianmcook:
Thanks! Regarding why I need to create many empty documents: I work with colleagues to build training courses containing many short videos and readings. We use Google Docs to author the video scripts and readings (because we need collaborative real-time editing). We first create a master course outline in a Google Sheet, then create separate empty Google Docs for all the videos and readings, then link from the master Sheet to every Doc, then finally do the authoring and editing in the Docs. It saves us many hours to automate the creation, naming, and linking of all those empty Google Docs, and googledrive and googlesheets4 have been invaluable for this. Thank you!

PR #266, opened 2019-08-10
Introduce the general idea of overwrite NA/TRUE/FALSE

### LINKED ISSUE #230: Overwrite control
Two minor things:

* Can the parameter to the local file for `drive_update()` be harmonized with the parameter for `drive_upload()` ? Currently the former is called "file" and the latter is called "path". Somewhat confusing...

* Would it be possible to have an "overwrite = TRUE" parameter in `drive_upload` ? Or alternatively enable a fallback option (parameter "create = TRUE") in `drive_update()` so that it creates files instead of raising an error if the file is not existing. Or combine both functions into one ?

jennybc:
Closes #230 

I decided that, if `overwrite` happens anywhere (such as in `drive_upload()`, it should happen for the entire set of functions that create new files.

`drive_put()` is also new.


jennybc:
@LucyMcGowan This is the biggest piece of work in googledrive since its release (other than the adoption of gargle, which didn't really add or change a feature *per se*).

I still need to add tests for this and make sure `overwrite` appears in some examples.

And I can fly solo here. But if you are in a position to look this over, it is always great to have another set of eyeballs. Even if you just read and comment, that's welcome.

LucyMcGowan:
I took a quick look, and this looks awesome! The only thing I noticed is we already had an overwrite parameter for `drive_download()` that is a bit different since it can just be `TRUE/FALSE` (since the local file systems aren't as wonky as the Drive ones 😂) - do you think we should update that wording to match-ish (with the exception of `NA`) or at least point out the distinction?

jennybc:
Thanks @LucyMcGowan, that's a good point. I added a "local" qualifier there, to serve as a reminder of where the overwrite could occur.

PR #273, opened 2019-09-02
Remove any duplicated nodes #272

### LINKED ISSUE #272: Intermittent duplicates in drive_ls()
I get intermittent errors when using `drive_ls()` on one of my folders.

```R
library(googledrive)
drive_auth(email = "my_email")
drive_ls("~/some/folder/")
#> Error in add_id_path(nodes, root_id = root_id, leaf = leaf): 
#> !anyDuplicated(nodes$id) is not TRUE
```

For some strange reason, this fails most of the time, but sporadically succeeds, in the same folder!

PR to follow.

andrie:
This fixes #272


jennybc:
I have applied a fix elsewhere, closer to what I think might be the root cause? e56b3f5d2e7b3a421115a1858df7456857478f5a

PR #282, opened 2019-11-07
Add LICENSE.md for tidyverse paper

batpigandme:
Adds license as markdown for tidyverse paper review (n.b. I copied the name wording from the plaintext license, using multiple "and"s, rather than semi-colon delimiting).

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -20,3 +20,4 @@
 ^vignettes/articles$
 ^revdep$
 ^index\.md$
+^LICENSE\.md$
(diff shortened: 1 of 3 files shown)


jennybc:
I actually now believe I should not be listed as a copyright holder, since all of my work on googledrive has occurred while working for RStudio.

@Hadley What about @LucyMcGowan? Her work on this occurred while an RStudio intern. Should she be listed as a copyright holder or should it just be RStudio?

hadley:
If she was paid by RStudio to work on it, then RStudio would own the copyright.

PR #296, opened 2020-01-31
Update link, add note on invalidity of inputs within examples.

### LINKED ISSUE #289: Refresh `drive_get()` docs
* [ ] Proofread title, description, and parameters 
* [ ] Proofread comments in examples 
* [ ] Ensure comment headings used to divide into scannable sections 
* [ ] Flag examples that seem overly complex for further review 
* [ ] Switch out mtcars/iris for something more interesting, or a smaller dataset constructed specifically to illustrate one point. 
* [ ] Review size of example outputs to ensure that you can easily see important rows and cols. 
* [ ] Datasets should always be tibbles to avoid use of `as_tibble()` distracting from the main point

smingerson:
- [x] Proofread title, description, and parameters
- [x] Proofread comments in examples
- [x] Ensure comment headings used to divide into scannable sections
- [x] Flag examples that seem overly complex for further review
- [x] ~~Switch out mtcars/iris for something more interesting, or a smaller dataset constructed specifically to illustrate one point.~~
- [x] Review size of example outputs to ensure that you can easily see important rows and cols.
- [x] ~~Datasets should always be tibbles to avoid use of as_tibble() distracting from the main point~~

Checks are what I did, strikethroughs do not seem relevant to the issue. 

### Summary of changes:

* Update a link in `@seealso` to point to the `files.get` endpoint link instead of `files.update`. This seems to be correct because in the preceding text it references files.get/list, not files.update/list.

* Add a note at the top of the examples indicating many will return 0-row tibbles or error unless the function inputs are replaced with paths or ids which exist on the user's Drive.

### Questions:

The return value of the function is shared through a template, which says "An object of class [`dribble`], a tibble with one row per item." It seems reasonable to mention somewhere that if no items are found, it will return a 0-row tibble (and not error/warn). However, I'm unsure if other functions which use the template share this behavior. It looks to be most (all?) of the functions use this template for the return. What's action would you recommend?

Closes #289

### DIFF
--- a/R/drive_get.R
+++ b/R/drive_get.R
@@ -34,15 +34,16 @@
 #'
 #' @examples
 #' \dontrun{
-#' ## Note many of these will error or return 0-row tibbles unless
-#' ## you replace the inputs with paths or ids which exist on your
-#' ## Drive.
-#'
 #' ## get info about your "My Drive" root folder
 #' drive_get("~/")
 #' ## the API reserves the file id "root" for your root folder
 #' drive_get(id = "root")
 #' drive_get(id = "root") %>% drive_reveal("path")
+#'
+#' ## The examples below are indicative of correct syntax.
+#' ## But note these will generally result in an error or a
+#' ## 0-row dribble, unless you replace the inputs with paths
+#' ## or file ids that exist in your Drive.
 #'
 #' ## multiple names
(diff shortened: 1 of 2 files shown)


jennybc:
Thanks!

Yeah I don't see a better way to handle the zero-row dribble thing at this point.

PR #333, opened 2021-04-14
Deprecate `verbose`; introduce googledrive_quiet option

jennybc:
I'm happy with how this works, but am inviting review of a very specific aspect (I comment this inline): how I determine the name of the calling function.

Big picture: googledrive had a `verbose` argument in 26 user-facing functions. I'm removing that, in favour of an option. I don't think many people actually need or want to fiddle with verbosity at the level of individual function calls. Very few users want to fiddle with verbosity at all, I imagine.

I want to wrap `lifecycle::deprecate_warn()` (i.e. use a shared `id`) and then call it from all 26 functions.

`warn_for_verbose()` catches the rare use of `drive_something(..., verbose = FALSE)`, warns, suggests what to do in future, and calls `local_drive_quiet()`.


R/drive_cp.R
```diff
@@ -74,7 +74,11 @@ drive_cp <- function(file,
                      name = NULL,
                      ...,
                      overwrite = NA,
-                     verbose = TRUE) {
+                     verbose = deprecated()) {
+  if (lifecycle::is_present(verbose)) {
```
jennybc, line 78: This is the first of a gazillion such insertions, which is why I want `warn_for_verbose()` to figure out the function from which it was called.

R/drive_cp.R
```diff
@@ -107,9 +111,7 @@ drive_cp <- function(file,
 
   out <- as_dribble(list(proc_res))
 
-  if (verbose) {
-    new_path <- paste0(append_slash(path$name), out$name)
-    message_glue("\nFile copied:\n  * {file$name} -> {new_path}")
-  }
+  new_path <- paste0(append_slash(path$name), out$name)
+  message_glue("\nFile copied:\n  * {file$name} -> {new_path}")
```
jennybc, line 115: FYI: `message_glue()` and friends will be replaced next, with cli-using functions. In a separate unit of work.

R/utils-ui.R
```diff
@@ -1,6 +1,118 @@
...
+    return(invisible())
+  }
+
+  called_from <- sys.parent(1)
+  if (called_from == 0) {
+    # called from global env, presumably in a test or during development
+    called_as <- "some_googledrive_function()"
+  } else {
+    called_as <- deparse(sys.call(called_from))
+  }
```
jennybc, line 19: Is this the right way to do this? I don't think there is any non-deprecated rlang function that is more suited to this, but I'd love your confirmation of that @lionel- .

Staying true to my brand, I do want to be able to call this function from `globalenv()`, just because it's convenient. But "in production", it's always called from some user-facing googledrive function.

lionel-: This looks good. However I wouldn't deparse the call and then reparse it with a regexp because that is brittle and more convoluted. Instead you could:

- Check that the result of `sys.call()` is a call (just to be defensive, in some cases/versions of R it sometimes returns a literal promise object) and that the first element of the call is a symbol (it could be another call or an inlined function).
- If the checks agree, convert the first element to string and use that. Otherwise use the default string.

lionel- (commented):
Perhaps we should add `caller_name()` to rlang? It would return `NULL` if no name can be inferred. Or perhaps it'd take a mandatory `default` argument to force the caller to deal with that situation.

R/utils-ui.R
hadley: Or match the lifecycle syntax a little more, and require the caller to specify the function name:

```R
warn_for_verbose(verbose, "my_function")
```

If you did want keep automatically guessing the function name, you could also lean into the call data structure and use that to generate the whole string:

```R
called_from <- sys.parent(1)
if (called_from == 0) {
  # called from global env, presumably in a test or during development
  called_as <- quote(some_googledrive_function())
} else {
  called_as <- sys.call(called_from)
}

called_as[[2]] <- quote(verbose)
caller <- as.character(called_as)
```

R/drive_find.R
```diff
@@ -126,14 +126,17 @@ drive_find <- function(pattern = NULL,
...
-                       verbose = TRUE,
+                       verbose = deprecated(),
                        team_drive = deprecated()) {
   if (!is.null(pattern) && !(is_string(pattern))) {
     stop_glue("`pattern` must be a character string.")
   }
   stopifnot(is_toggle(trashed))
   stopifnot(is.numeric(n_max), n_max >= 0, length(n_max) == 1)
 
+  if (lifecycle::is_present(verbose)) {
```
hadley, line 137: I'd suggest also including the `is_present()` check in `warn_for_verbose()`

jennybc:
> Perhaps we should add `caller_name()` to rlang? It would return `NULL` if no name can be inferred. Or perhaps it'd take a mandatory default argument to force the caller to deal with that situation.

I was a bit surprised that there wasn't an rlang function that makes this easier. I have encountered this need a couple times before and the lifecycle functions are a great example: you're trying to throw an actionable or educational error that tells people exactly how they've goofed up.

usethis uses the low-tech "who's asking?" approach in several of its checkers, e.g.:

https://github.com/r-lib/usethis/blob/0d6522474e35c78391a0b54dc12d63a63a5ce05a/R/proj.R#L198-L211

The simplicity of that is probably a better idea, in general.

But it was interesting to explore the auto-discovery method here.

PR #342, opened 2021-04-27
Fix for garbled multibyte Google Sheets title in drive_upload

hideaki:
## Issue to resolve
On Windows, when uploading a CSV file as a Google Sheets spreadsheet with a sheet title that includes multibyte characters, the title of the resulting uploaded spreadsheet is garbled.

e.g.
```
googledrive::drive_upload(csv_filepath, "Title with multibyte - マルチバイトのタイトル", type = "spreadsheet", overwrite = TRUE)
```

## Cause
This is because of unnecessary encoding translation done by writeLines(), which writes the JSON content to the temporary file for subsequent HTTP upload request. The content of the written JSON file gets garbled. 

## Fix
To avoid the unnecessary encoding translation, I added useBytes = TRUE argument to the writeLines call.
Reference: https://stackoverflow.com/questions/10675360/utf-8-file-output-in-r

## Test
Tested that this resolves the garbling issue on Windows, and also drive_upload keeps working on Mac.

## R CMD check results
```
-- R CMD check results -------------------------------- googledrive 1.0.1.2 ----
Duration: 1m 3.9s

> checking for missing documentation entries ... WARNING
  Undocumented code objects:
    'drive_set_token'
  All user-level objects in a package should have documentation entries.
  See chapter 'Writing R documentation files' in the 'Writing R
  Extensions' manual.

> checking R/sysdata.rda ... OK
   WARNING
  'qpdf' is needed for checks on size reduction of PDFs

> checking package dependencies ... NOTE
  Package suggested but not available for checking: 'spelling'

> checking for future file timestamps ... NOTE
  unable to verify current time

0 errors v | 2 warnings x | 2 notes x
Error: R CMD check found WARNINGs
Execution halted
```

### DIFF
--- a/R/drive_upload.R
+++ b/R/drive_upload.R
@@ -129,7 +129,8 @@ drive_upload <- function(media,
   meta_file <- withr::local_file(
     tempfile("drive-upload-meta", fileext = ".json")
   )
-  writeLines(jsonlite::toJSON(params), meta_file)
+  ## useBytes is set to avoid unnecessary encoding translation on Windows.
+  writeLines(jsonlite::toJSON(params), meta_file, useBytes = TRUE)
   ## media uploads have unique body situations, so customizing here.
   request$body <- list(
     metadata = httr::upload_file(


jennybc:
Thanks for identifying this problem and providing a clear reprex.

As you'll see, I've fixed it myself, using an approach that is more consistent with things we do elsewhere (in tidyverse / r-lib), which is ultimately more maintainable for me.

If you are motivated and take a look at the TODO section here (https://github.com/tidyverse/googledrive/pull/344/files#diff-ffc0baedb4d0cb3963ea33c48c1c3b5c1ec2c4cbea95e6438cc0929a830715dd), I'm curious:

Have you ever tried to use `drive_upload()` *without specifying `path` or `name`* and instead just allowing the Drive file's name to be determined from the base name of `media`? And have you done that with a file whose name contains CJK characters? I would like to make that work, but I think it would actually require a change in curl (as in the R package). But I also wonder if maybe it already "just works" for someone who works with CJK regularly on Windows and, therefore, may have a different configuration than I do on Windows. Or maybe it "just works" after I merge my PR? If you have any observations on this, I'd be interested to hear more.

jennybc:
Also, can you show me your output for `l10n_info()`?

PR #347, opened 2021-05-05
Update dplyr compat, add vctrs compat

### LINKED ISSUE #338: Update dplyr compatibility

jennybc:
Closes #338


jennybc:
I'd appreciate a quick look from you @DavisVaughan. I have a couple of questions and want to comment on a few places where things were different from the examples you've written up. I think the main difference is that googledrive only Suggests dplyr (vs. Imports).

R/compat-dplyr.R
```diff
@@ -0,0 +1,39 @@
+dribble_maybe_reconstruct <- function(data, template) {
+  if (dribble_is_reconstructable(data)) {
+    # in workflowsets, davis uses new_workflow_set0(x) here
+    new_dribble(data)
+  } else {
+    # davis says:
+    # in dribble_maybe_reconstruct(), rather than as_tibble() you may want to
+    # use new_tibble(). I think there is a chance that the input could actually
+    # just be a list, rather than a data frame, so really you are just adding
+    # the appropriate data frame specific attributes back on the list.
```
jennybc, line 10: @DavisVaughan Are you saying that some internal machinations inside dplyr or vctrs could lead to an input that's a list? And, if that's the case, can I rely on the list components having the same length? I see that you check for a list in workflowsets, but you don't check for same length.

R/googledrive-package.R
```diff
@@ -42,6 +42,9 @@
 #' @importFrom lifecycle deprecated
 #' @importFrom rlang :=
 #' @importFrom rlang %||%
+#' @importFrom vctrs vec_cast
+#' @importFrom vctrs vec_ptype2
+#' @importFrom vctrs vec_restore
```
jennybc, line 47: I don't (can't) import `dplyr::dplyr_reconstruct()`, since dplyr is not in Imports. In your examples I studied, you import the entire vctrs namespace, so this is also a bit different.

R/zzz.R
```diff
@@ -0,0 +1,24 @@
...
+    )
+  }
+
+  if (is_installed("dplyr", version = "1.0.0")) {
+    vctrs::s3_register(
+      "dplyr::dplyr_reconstruct",
+      "dribble",
+      method = dribble_maybe_reconstruct
+    )
+  }
```
jennybc, line 21: I have to do this since dplyr is only in Suggests. I've got this condition, because I'm willing to only support dplyr compatibility from v1.0.0 on. And this use of `version` is, in fact, why I depend on dev rlang in this PR (but I will need this for separate rlang+cli reasons anyway).

tests/testthat/test-compat-dplyr.R
```diff
@@ -0,0 +1,263 @@
+# based on https://github.com/tidymodels/workflowsets/blob/main/tests/testthat/test-compat-dplyr.R
```
jennybc, line 1: The tests are very much yours, with some tweaks because I don't have the row uniqueness issue you have in workflowsets. I also migrated some existing tests I had of a few other dplyr verbs.

DESCRIPTION
```diff
@@ -29,15 +29,16 @@ Imports:
     lifecycle,
     magrittr,
     purrr (>= 0.2.3),
-    rlang (>= 0.3.1),
+    rlang (>= 0.4.11.9000),
     tibble (>= 2.0.0),
     utils,
     uuid,
+    vctrs,
```
jennybc, line 36: Should I be putting a minimum version on vctrs?

DavisVaughan: I would use at least 0.3.0, as that was the version where we changed the ptype2/cast system to what you are using here

R/compat-dplyr.R
DavisVaughan: Yea, exactly that. I can't remember exactly when it could occur, but, if it did happen, the unclassed list would most likely come through as `x` in your `vec_restore()` method.

The components should all still be guaranteed to have the same length, so checking that shouldn't be required.

R/compat-dplyr.R
```diff
@@ -0,0 +1,39 @@
...
+
+dribble_is_reconstructable <- function(data) {
+  # continuing davis's comments from above:
+  # ... that might also mean you should remove the is.data.frame() check that is
+  # there, and instead just check the structure of the object (col names, types,
+  # etc).
+  #
+  # in workflowsets, he uses has_required_container_type(),
+  # which is just rlang::is_list()
+  is.data.frame(data) &&
```
DavisVaughan, line 25: I think I was suggesting that this become `is_list(data)`

In the case that `data` somehow became a bare list, it could still be reconstructable to a dribble if all of your other conditions pass

R/compat-dplyr.R
```diff
@@ -0,0 +1,39 @@
...
+    has_dribble_cols(data) &&
+    has_dribble_coltypes(data) &&
+    has_drive_resource(data)
+
+  # TODO: I suspect I should not accept NAs in the id column
+  # maybe also name?
+}
+
+new_tibble0 <- function(x, ..., class = NULL) {
```
DavisVaughan, line 34: I have long wanted `new_tibble()` to drop attributes so we don't need this https://github.com/tidyverse/tibble/pull/769

R/compat-vctrs.R
```diff
@@ -0,0 +1,102 @@
...
+# which should always result in a new dribble, as long as
+# `df_ptype2()` can compute a common type.
+#
+# Combining a dribble with a tibble/data.frame will only ever happen if
+# the user calls `vec_c()` or `vec_rbind()` with one of each of those inputs.
+# Although I could probably make this work, it feels pretty weird and exotic
+# and I think that user, if they even exist, should just turn that "other"
+# tibble/data.frame into a dribble first.
+# So I'll follow workflowsets and not attempt to return dribble for these
+# combinations.
```
DavisVaughan, line 29: I find that keeping it as simple as possible (as you've done here) is much more rewarding. Trying to be too "smart" always ended up being extremely buggy and fragile.

R/googledrive-package.R
DavisVaughan: I'm very liberal with always `@import`-ing both vctrs and rlang in anything I work on, but this is completely fine

R/zzz.R
DavisVaughan: That's fine, its very similar to what we do in tune, since at the time we had to conditionally register depending on the dplyr version that was being used (even though it was in Imports) https://github.com/tidymodels/tune/blob/b0e83a776df5779ee325938b217ce7a5192b715f/R/zzz.R#L78

I have a faint memory of `is_installed()` not working correctly inside `.onLoad()`, but can't remember exactly where that came up. But if it isn't bothering you, I wouldn't worry about it.

tests/testthat/test-compat-dplyr.R
```diff
@@ -0,0 +1,263 @@
...
+
```
DavisVaughan, line 2: This is one of the very _very_ few places where I moved away from completely self-contained tests in favor of skipping the entire file, since every single test here requires dplyr 1.0.0
https://github.com/tidymodels/tune/blob/b0e83a776df5779ee325938b217ce7a5192b715f/tests/testthat/test-compat-dplyr.R#L2

DavisVaughan (approved):
It looks great!

PR #359, opened 2021-05-24
Keep checking on Windows + 3.6; reinstate the cache workaround

jennybc:


jennybc:
Saving for posterity, here's what's different re: system PATH before and after installing R:

```
> dat %>% rowid_to_column() %>% filter(mismatch)
# A tibble: 7 x 4
  rowid bef                      aft                              mismatch
  <int> <chr>                    <chr>                            <lgl>   
1     1 /mingw64/bin             /c/R/bin                         TRUE    
2     2 /usr/bin                 /mingw64/bin                     TRUE    
3     3 /c/Users/runneradmin/bin /usr/bin                         TRUE    
4    21 /bin                     /c/Program Files/Git/bin         TRUE    
5    60 /cmd                     /c/Program Files/Git/cmd         TRUE    
6    61 /mingw64/bin             /c/Program Files/Git/mingw64/bin TRUE    
7    62 /usr/bin                 /c/Program Files/Git/usr/bin     TRUE    
```

Basically `/c/R/bin/` has been appended at the front.

`/c/Users/runneradmin/bin` has gone missing.

I guess the other stuff might be from doing something like `normalizePath()`? In fact, maybe that's also what happened with `/c/Users/runneradmin/bin`, i.e. maybe it doesn't point to anything?

jennybc:
I did this directly in `master`, but this was sort of interesting.

PR #408, opened 2023-01-13
Modernize the warn_for_verbose() helper

jennybc:
The motivation for this update is that I will probably feature this as a "deprecation helper" in R Packages. I.e. the type of thing one might do if a deprecation affects a lot of code, such as when googledrive deprecated the `verbose` argument present in virtually every single function in favour of an option.

The helper has been here for quite a while, so I'm updating it to remove some DIY moves and be more idiomatic with modern lifecycle.

How it gets used in almost every function in googledrive:

``` r
drive_create <- function(name,
                         path = NULL,
                         type = NULL,
                         ...,
                         overwrite = NA,
                         verbose = deprecated()) {
  warn_for_verbose(verbose)
  ...
}
```

### DIFF
--- a/R/utils-ui.R
+++ b/R/utils-ui.R
@@ -199,9 +199,10 @@ message <- function(...) {
 warn_for_verbose <- function(verbose = TRUE,
                              env = caller_env(),
                              user_env = caller_env(2)) {
-  # this is about whether `verbose` was present in the **user's** call to the
-  # calling function
-  # don't worry about the `verbose = TRUE` default in warn_for_verbose()
+  # This function is not meant to be called directly, so don't worry about its
+  # default of `verbose = TRUE`.
+  # In authentic, indirect usage of this helper, this picks up on whether
+  # `verbose` was present in the **user's** call to the calling function.
   if (!lifecycle::is_present(verbose) || isTRUE(verbose)) {
     return(invisible())
   }
@@ -218,7 +219,7 @@ warn_for_verbose <- function(verbose = TRUE,
     always = identical(env, global_env()),
     id = "googledrive_verbose"
   )
-  # only set the option during organic, indirect usage
+  # only set the option during authentic, indirect usage
   if (!identical(env, global_env())) {
     local_drive_quiet(env = env)
   }
--- a/R/utils-ui.R
+++ b/R/utils-ui.R
@@ -216,8 +216,7 @@ warn_for_verbose <- function(verbose = TRUE,
       "googledrive's `verbose` argument will be removed in the future."
     ),
     user_env = user_env,
-    always = identical(env, global_env()),
-    id = "googledrive_verbose"
+    always = identical(env, global_env())
   )
   # only set the option during authentic, indirect usage
   if (!identical(env, global_env())) {
--- a/R/utils-ui.R
+++ b/R/utils-ui.R
@@ -207,16 +207,25 @@ warn_for_verbose <- function(verbose = TRUE,
     return(invisible())
   }
 
+  fc <- frame_call(env)
+  caller <- if (is.null(fc)) NULL else call_name(fc)
+  if (is.null(caller)) {
+    what = I("The `verbose` argument")
+  } else {
+    what = glue("{caller}(verbose)")
+  }
+
   lifecycle::deprecate_warn(
     when = "2.0.0",
-    what = I("The `verbose` argument"),
+    what = what,
     details = c(
       "Set `options(googledrive_quiet = TRUE)` to suppress all googledrive messages.",
       "For finer control, use `local_drive_quiet()` or `with_drive_quiet()`.",
       "googledrive's `verbose` argument will be removed in the future."
     ),
     user_env = user_env,
-    always = identical(env, global_env())
+    always = identical(env, global_env()),
+    id = "googledrive_verbose"
   )
   # only set the option during authentic, indirect usage
   if (!identical(env, global_env())) {
--- a/R/utils-ui.R
+++ b/R/utils-ui.R
@@ -196,36 +196,31 @@ message <- function(...) {
     instead of {.fun message}")
 }
 
-warn_for_verbose <- function(verbose = TRUE, env = caller_env()) {
+warn_for_verbose <- function(verbose = TRUE,
+                             env = caller_env(),
+                             user_env = caller_env(2)) {
   # this is about whether `verbose` was present in the **user's** call to the
   # calling function
-  # don't worry about the `verbose = TRUE` default here
+  # don't worry about the `verbose = TRUE` default in warn_for_verbose()
   if (!lifecycle::is_present(verbose) || isTRUE(verbose)) {
     return(invisible())
   }
 
-  called_from <- sys.parent(1)
-  if (called_from == 0) {
-    # called from global env, presumably in a test or during development
-    caller <- "some_googledrive_function"
-  } else {
-    called_as <- sys.call(called_from)
-    if (is.call(called_as) && is.symbol(called_as[[1]])) {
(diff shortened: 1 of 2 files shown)


R/utils-ui.R
```diff
@@ -196,36 +196,31 @@ message <- function(...) {
...
-    # called from global env, presumably in a test or during development
-    caller <- "some_googledrive_function"
-  } else {
-    called_as <- sys.call(called_from)
-    if (is.call(called_as) && is.symbol(called_as[[1]])) {
-      caller <- as.character(called_as[[1]])
-    } else {
-      caller <- "some_googledrive_function"
-    }
-  }
```
jennybc, line 218: The aforementioned "DIY moves" in which I determined the name of the calling function. In this update, I am letting go of this in favour of a simpler function. I'll probably remove `verbose` and this helper entirely in the next release of googledrive anyway.

R/utils-ui.R
```diff
@@ -196,36 +196,31 @@ message <- function(...) {
...
     when = "2.0.0",
-    what = glue("{caller}(verbose)"),
+    what = I(cli::format_inline("The {.arg verbose} argument")),
     details = c(
       "Set `options(googledrive_quiet = TRUE)` to suppress all googledrive messages.",
       "For finer control, use `local_drive_quiet()` or `with_drive_quiet()`.",
       "googledrive's `verbose` argument will be removed in the future."
     ),
+    user_env = user_env,
+    always = identical(env, global_env()),
```
jennybc, line 218: I do like to be able to call this after `load_all()` in the global env and see the message.

jennybc:
@Hadley I'm asking for your review somewhat authentically, but mostly because I'm probably going to feature this in R Packages as an example of a deprecation helper.

R/utils-ui.R
```diff
@@ -196,36 +196,31 @@ message <- function(...) {
...
-    if (is.call(called_as) && is.symbol(called_as[[1]])) {
-      caller <- as.character(called_as[[1]])
-    } else {
-      caller <- "some_googledrive_function"
-    }
-  }
   lifecycle::deprecate_warn(
     when = "2.0.0",
-    what = glue("{caller}(verbose)"),
+    what = I(cli::format_inline("The {.arg verbose} argument")),
```
hadley, line 211: I think it would be fine to just use ``"The `verbose` argument"``; using cli in this way is rather unusual so I wouldn't want to feature it too prominently. 

I'd almost prefer keeping `caller` and then doing `what = paste0(caller, "(verbose)")`.

R/utils-ui.R
hadley: That makes sense, but again if you want to feature it in R packages, I'd think you'd be better off sticking to standard usage.

R/utils-ui.R
jennybc: But my impression is that I won't know `caller` without my DIY gymnastics, which I'm removing. I'm not aware of any clean way to get that with simple usage of lifecycle. Do you think I'm missing something?

jennybc: (I went ahead and removed the use of `cli::format_inline()`.)

R/utils-ui.R
jennybc: I can redact that part in the book, meaning the specification of `always =`.

R/utils-ui.R
hadley: Yeah, I meant keep `caller` (or make it an explicit argument to pass along)

jennybc: I don't think we want to subject R Packages readers to my code for learning `caller`, so the version I put in the book will be a simple statement about `verbose`. But maybe, before I actually merge this, I'll put the `caller` stuff back.

This is PR is also to make sure that whatever I put in the book actually works, even if it's not what I ultimately use.

jennybc: I thought about the explicit `caller` at the very start, but it would be pretty annoying in this particular case.

jennybc:
Since I reminded myself that what I'm doing works and why it works, I think I'll park this little demo here for posterity.

``` r
library(lifecycle)

warn_for_verbose <- function(verbose = TRUE) {
  if (is_present(verbose)) {
    cat("`is_present(verbose)` is TRUE\n")
  } else {
    cat("`is_present(verbose)` is FALSE\n")
  }
}

# direct usage
warn_for_verbose()
#> `is_present(verbose)` is TRUE
warn_for_verbose(TRUE)
#> `is_present(verbose)` is TRUE
warn_for_verbose(FALSE)
#> `is_present(verbose)` is TRUE

f <- function(verbose = deprecated()) {
  warn_for_verbose(verbose)
  "hi"
}

# authentic, indirect, helper-style usage
f()
#> `is_present(verbose)` is FALSE
#> [1] "hi"
f(TRUE)
#> `is_present(verbose)` is TRUE
#> [1] "hi"
f(FALSE)
#> `is_present(verbose)` is TRUE
#> [1] "hi"
```

<sup>Created on 2023-01-13 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

R/utils-ui.R
```diff
@@ -196,36 +196,31 @@ message <- function(...) {
...
-    what = glue("{caller}(verbose)"),
+    what = I("The `verbose` argument"),
     details = c(
       "Set `options(googledrive_quiet = TRUE)` to suppress all googledrive messages.",
       "For finer control, use `local_drive_quiet()` or `with_drive_quiet()`.",
       "googledrive's `verbose` argument will be removed in the future."
     ),
-    id = "googledrive_verbose"
+    user_env = user_env,
+    always = identical(env, global_env())
```
lionel-, line 219: hmm, should this be the default behaviour in lifecycle? I.e. if the deprecation is called directly, we always warn. It's supposed to be called from a deprecated feature, and if we see it's called from the global env it is likely during dev.

R/utils-ui.R
```diff
@@ -196,36 +196,31 @@ message <- function(...) {
...
-    if (is.call(called_as) && is.symbol(called_as[[1]])) {
-      caller <- as.character(called_as[[1]])
-    } else {
-      caller <- "some_googledrive_function"
-    }
-  }
   lifecycle::deprecate_warn(
     when = "2.0.0",
-    what = glue("{caller}(verbose)"),
+    what = I("The `verbose` argument"),
```
lionel-, line 212: Why this change? Is it to simplify the pattern for the book?

I think this is where the mini-DSL for specifying deprecated features falls apart, i.e. it's nice for direct usage but causes complications when programmed.

Ideally we'd keep mentioning the function where the argument is deprecated. And ideally we wouldn't have to specify it and automatically infer it from `env`.

So I guess to help programmatic usage we'd need some sort of ctor like `what = deprec(arg = "my_arg")`. You could supply `fn = ` but it would be automatically inferred from `env` if not supplied.

lionel-:
To retrieve the caller, this could be helpful:

```r
call_name(frame_call(env))
```

`frame_call()` retrieves the `sys.call()` of a frame environment

`call_name()` is from the first versions of rlang. I haven't used it much but looks like it may be useful here. It can return NULL so this should be guarded within a stopifnot.

R/utils-ui.R
jennybc: > Why this change? Is it to simplify the pattern for the book?

Yes, exactly.

R/utils-ui.R
lionel-: https://github.com/r-lib/lifecycle/issues/157

jennybc:
I was able to use @lionel-'s advice to bring back `caller`, but using rlang. I think this is pretty good now. The book has a simplified version that doesn't attempt to infer the calling function, which I think is the right call.

PR #416, opened 2023-03-17
Make the method's signature more closely match the generic's

jennybc:
Purposefully trying just `...` first to see what happens


jennybc:
The `with-auth` failure is unrelated to this.

PR #434, opened 2023-06-27
Optionally skip tests requiring 'mockr'

MichaelChirico:
It would be simpler to add the `skip()` inside the helper:

https://github.com/tidyverse/googledrive/blob/b87d89ec5368e7c9800978ae2a8af71a34a20b54/tests/testthat/helper.R#L16-L18

But the tests would have to be rewritten a bit, e.g.

https://github.com/tidyverse/googledrive/blob/b87d89ec5368e7c9800978ae2a8af71a34a20b54/tests/testthat/test-drive_get_path.R#L68-L74

Would probably become:

```r
with_mock(
  root_id = function() "",
  expect_equal(resolve_paths(as_dribble(x), ancestors)$path, "e")
)
```

That format of test looks more natural to me anyway, but filing as is now pending feedback.

### DIFF
--- a/tests/testthat/test-drive_get_path.R
+++ b/tests/testthat/test-drive_get_path.R
@@ -23,6 +23,7 @@ test_that("get_last_path_part() works", {
 })
 
 test_that("resolve_paths() works, basic scenarios", {
+  skip_if_not_installed("mockr")
   # a -- b -- c -- d
   # ??? -- e
   dr_folder <-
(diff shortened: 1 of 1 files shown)


jennybc:
I've been removing mockr dependencies in favor of `testthat::local_mocked_bindings()`. Apparently I just haven't done so here yet. So I'm not going to take a change that touches this stuff in any other way; it's not worth fiddling with it.

MichaelChirico:
Makes sense! I'll leave this open for now in case it serves as a reminder, but feel free to close at any time.

PR #458, opened 2024-08-08
Use skip_if() over custom approach

MichaelChirico:
This is functionally basically identical, but it's better to use the specific function since it's there.


jennybc:
Thanks!

